Issue #19, opened 2023-03-20
[FR] shrinkWrap option for AppFlowyEditor


Xazin:
To enable the Editor to be inside other ScrollViews, we can add a shrinkWrap option that defaults to false (scrollable).

This is especially useful when you want to have a view-only mode for the Editor, and for it to be inside another ScrollView.

PR #20, opened 2023-03-20
feat: shrinkWrap option for AppFlowyEditor

Xazin:
Relates: #19

### DIFF
--- a/lib/src/core/transform/operation.dart
+++ b/lib/src/core/transform/operation.dart
@@ -235,8 +235,7 @@ class UpdateTextOperation extends Operation {
 
 // TODO(Lucas.Xu): refactor this part
 Path transformPath(Path preInsertPath, Path b, [int delta = 1]) {
-  if ((preInsertPath.length > b.length) ||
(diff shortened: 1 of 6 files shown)


lib/src/core/transform/operation.dart
```diff
@@ -235,28 +235,31 @@ class UpdateTextOperation extends Operation {
 
 // TODO(Lucas.Xu): refactor this part
 Path transformPath(Path preInsertPath, Path b, [int delta = 1]) {
-  if (preInsertPath.length > b.length) {
-    return b;
-  }
-  if (preInsertPath.isEmpty || b.isEmpty) {
+  if ((preInsertPath.length > b.length) ||
+      (preInsertPath.isEmpty || b.isEmpty)) {
```
Xazin, line 239:
```suggestion
  if (preInsertPath.length > b.length || preInsertPath.isEmpty || b.isEmpty) {
```
Don't know why I consolidated it like that, the above suggestion is as good.

Issue #16, opened 2023-03-16
[FR] Improve test code coverage.


LucasXu0:
Now, the code coverage for the editor is 77%. We aim to 100%.

Reference
* https://app.codecov.io/github/AppFlowy-IO/appflowy-editor

LucasXu0:
<img width="1621" alt="Screenshot 2023-03-16 at 09 57 05" src="https://user-images.githubusercontent.com/11863087/225489898-4c042b66-8269-483f-811f-0a019847b375.png">

PR #21, opened 2023-03-20
tests: improve test coverage for commands

Xazin:
100% coverage for `text_commands.dart` and 80% coverage for `command_extension.dart`.

Relates: #16 

**Note: There were some methods in the extensions which took in an EditorState, it seems quite weird to me, but since it's already an extension on EditorState, I removed that. Let me know if I shouldn't touch it, and I will put it back.**


LucasXu0:
LGTM. Thanks for contributing.

PR #28, opened 2023-03-22
feat: alt + arrow key to move cursor one word

Xazin:
Relates: https://github.com/AppFlowy-IO/AppFlowy/issues/2071

### DIFF
--- a/test/service/shortcut_event/shortcut_event_test.dart
+++ b/test/service/shortcut_event/shortcut_event_test.dart
@@ -120,7 +120,7 @@ void main() async {
           event.updateCommand(
             windowsCommand: 'ctrl+arrow right',
             linuxCommand: 'ctrl+arrow right',
-            macOSCommand: 'ctrl+arrow rightt',
+            macOSCommand: 'ctrl+arrow right',
           );
         }
       }
--- a/test/service/shortcut_event/shortcut_event_test.dart
+++ b/test/service/shortcut_event/shortcut_event_test.dart
@@ -61,16 +61,17 @@ void main() async {
       for (final event in builtInShortcutEvents) {
         if (event.key == 'Move cursor begin') {
           event.updateCommand(
-            windowsCommand: 'alt+arrow left',
-            linuxCommand: 'alt+arrow left',
-            macOSCommand: 'alt+arrow left',
+            windowsCommand: 'ctrl+arrow left',
+            linuxCommand: 'ctrl+arrow left',
+            macOSCommand: 'ctrl+arrow left',
           );
         }
       }
+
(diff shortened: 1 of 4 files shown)


Xazin:
@LucasXu0 When you have time, can you guide me on how to resolve the issue with tests. I had to make changes because I took over the command that the failing tests were taking advantage of.

test/service/shortcut_event/shortcut_event_test.dart
```diff
@@ -117,15 +118,16 @@ void main() async {
...
+            linuxCommand: 'ctrl+arrow right',
+            macOSCommand: 'ctrl+arrow rightt',
           );
         }
       }
+
       await editor.pressLogicKey(
         LogicalKeyboardKey.arrowRight,
-        isAltPressed: true,
+        isControlPressed: true,
```
a-wallen, line 130: `appflowy_flutter\packages\appflowy_editor\lib\src\service\shortcut_event\key_mapping.dart` must not have an entry that handles this particular combination of keys which is why the test is failing.

test/service/shortcut_event/shortcut_event_test.dart
```diff
@@ -117,15 +118,16 @@ void main() async {
       for (final event in builtInShortcutEvents) {
         if (event.key == 'Move cursor end') {
           event.updateCommand(
-            windowsCommand: 'alt+arrow right',
-            linuxCommand: 'alt+arrow right',
-            macOSCommand: 'alt+arrow right',
+            windowsCommand: 'ctrl+arrow right',
+            linuxCommand: 'ctrl+arrow right',
+            macOSCommand: 'ctrl+arrow rightt',
```
LucasXu0, line 123: Hey, @Xazin. Double `t` here.

MayurSMahajan (commented):
In `test/service/shortcut_event/shortcut_event_test.dart` file when you are updating the command to ctrl+arrow left, in macOSCommand you want to use **meta**  instead of ctrl. On line 66: `macOsCommand: 'meta+arrow left',`

test/service/shortcut_event/shortcut_event_test.dart
Xazin: Thanks

test/service/shortcut_event/shortcut_event_test.dart
Xazin: This is actually done on purpose, I didn't actually touch anything that _should_ have changed this test from succeeding, but apparently a few lines above in the test, it "updates" a command, and the key events used to trigger it, which was ones I had used in the new command implemented in this feature, thus it failed due to the colliding commands, hence the change here.

Xazin:
Thanks @LucasXu0 - That was the issue ...

PR #30, opened 2023-03-23
fix: cannot edit image or see link menu if editorState is uneditable

Xazin:
There are two changes here:
- Stop showing action menu and edge resize on image node when editor is not editable
- Stop showing link menu on tapping a link if editor is not editable (if you click a link in non-editable mode, you open it.)


Xazin:
Force pushes because working on a Mac.. Had to hack the slash command to work.

Xazin:
MacOS Test run failed, but due to codecov not actual tests.

Xazin:
@LucasXu0 Can you rerun the failing action?

LucasXu0:
@Xazin Yes. I will merge it.

PR #45, opened 2023-04-06
fix: delete node does not propagate non null selection

### LINKED ISSUE #39: [Bug] replace does not work in certain use-cases
### Bug Description

This issue was discovered in appflowy, so I have used a usecase there to describe the issue.

As the title suggests, replace option when using OpenAi to generate a summary or fix-spelling does not work as expected in few use-cases, if response generated is multiline, last line overwrites the previous lines. This can cause much problem and confusion if user is working on large selections of text for correction or summary generation.

### How to Reproduce

1. Select some text on editor (likely a query to generate multiple lines from open ai)
2. select summarize and then replace the selected text.

### Expected Behavior

Selected text should be correctly replaced by the response generated by open-ai.

### Operating System

Linux

### AppFlowy Version(s)

Version 0.1.1 (1)

### Screenshots

https://user-images.githubusercontent.com/84044317/227204772-9077f453-4994-41da-9618-5ae431cac55e.mp4

### Additional Context

This issue was previously opened, resolved and merged in appflowy and I needed it to write correct integration tests for other dependent issue I was working on however, since dependencies there have been changed for appflowy editor, those fixes are now redundant. For more reference kindly take a look at these.

https://github.com/AppFlowy-IO/AppFlowy/pull/2100 (Now Redundant)
https://github.com/AppFlowy-IO/AppFlowy/pull/2107#issuecomment-1496291314

These changes will be tested in appflowy via integration tests.

squidrye:
fixes #39

### DIFF
--- a/lib/src/core/transform/transaction.dart
+++ b/lib/src/core/transform/transaction.dart
@@ -68,6 +68,9 @@ class Transaction {
   /// Deletes the [Node] in the document.
   void deleteNode(Node node) {
     deleteNodesAtPath(node.path);
+    if(beforeSelection != null){
+      afterSelection = beforeSelection;
+    }
   }
 
   /// Deletes the [Node]s in the document.
--- a/test/core/transform/transaction_test.dart
+++ b/test/core/transform/transaction_test.dart
@@ -202,5 +202,34 @@ void main() async {
       expect(textNodes[3].toPlainText(), 'ABC4');
       expect(textNodes[4].toPlainText(), 'ABC5 AppFlowy!');
     });
+
+    testWidgets('test selection preserves on node deletion in documents',
+        (tester) async {
+      TestWidgetsFlutterBinding.ensureInitialized();
+
+      final editor = tester.editor
+        ..insertTextNode('Welcome to AppFlowy!')
+        ..insertTextNode('Testing selection on this');
+
+      await editor.startTesting();
+      await tester.pumpAndSettle();
+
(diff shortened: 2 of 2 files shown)


lib/src/core/transform/transaction.dart
```diff
@@ -68,6 +68,9 @@ class Transaction {
   /// Deletes the [Node] in the document.
   void deleteNode(Node node) {
     deleteNodesAtPath(node.path);
+    if(beforeSelection != null){
```
LucasXu0, line 71:
```suggestion
    if (beforeSelection != null) {
```

LucasXu0:
Hey, @squidrye. Not every delete can assign the `beforeSelection` to the `afterSelection`. For example, 

**Before**
Node 1
|Node 2| <- delete it. || means the selection

**After**
Node 1| <- now, the `afterSelection` is not equal to the `beforeSelection`

squidrye:
Thanks for pointing that out, I'll see what can be done.

squidrye:
I have added a condition to check if node being deleted is currently selected node, if so after-selection will not be updated, but if node being deleted is not selected and some node in background is selected it will be propagated and maintained.

|Node 1|
Node 2 <- deleted then Node 1 will be still selected

|Node 1| <- deleted then afterselection will not be updated

Let me know if I am missing something else here.

LucasXu0:
Hey, @squidrye. Please format your code.

PR #50, opened 2023-04-08
feat: add local images from computer

rileyhawk1417:
Porting this plugin over from AppFlowy into AppFlowy-Editor, that allows you to add local image files in the document.


lib/src/render/image/image_node_widget.dart
```diff
@@ -138,6 +141,15 @@ class ImageNodeWidgetState extends State<ImageNodeWidget> with SelectableMixin {
...
   Widget _buildResizableImage(BuildContext context) {
+    final fileImage = Image.file(
+      //TODO:  Do not render the widget if null
+      File(widget.src),
+      width: _imageWidth == null ? null : _imageWidth! - _distance,
+      gaplessPlayback: true,
+      errorBuilder: (context, error, stackTrace) {
+        return _buildError(context);
+      },
+    );
```
Xazin, line 152:
```suggestion
    Image? image;
    switch (widget.type) {
      case 'file':
        final imageFile = File(widget.src);
      
        if (!imageFile.existsSync()) {
          return const SizedBox.shrink();
        }
      
        image = Image.file(
          imageFile,
          width: _imageWidth == null ? null : _imageWidth! - _distance,
          gaplessPlayback: true,
          errorBuilder: (context, error, stackTrace) {
            return _buildError(context);
          },
        );
      case 'network':
        image = Image.network(
          widget.src,
          width: _imageWidth == null ? null : _imageWidth! - _distance,
          gaplessPlayback: true,
          loadingBuilder: (context, child, loadingProgress) {
            if (loadingProgress == null ||
                loadingProgress.cumulativeBytesLoaded ==
                    loadingProgress.expectedTotalBytes) {
              return child;
            }

            return _buildLoading(context);
          },
          errorBuilder: (context, error, stackTrace) => _buildError(context),
        );
    }
```
Maybe even abstract this into a private method eg. `_buildImageFromType`

lib/src/render/image/image_node_widget.dart
```diff
@@ -161,7 +173,8 @@ class ImageNodeWidgetState extends State<ImageNodeWidget> with SelectableMixin {
 
     return Stack(
       children: [
-        networkImage,
+        if (widget.type.toString() == 'file') ...[fileImage],
+        if (widget.type.toString() == 'network') ...[networkImage],
```
Xazin, line 177: My above suggestion is because I don't think it's worth it to have both types of Image when only one is used.

lib/src/render/image/image_upload_widget.dart
```diff
@@ -84,8 +93,47 @@ class _ImageUploadMenuState extends State<ImageUploadMenu> {
...
 
+  void _pickFiles() async {
+    _resetState();
+    try {
+      _paths = (await FilePicker.platform.pickFiles(
+              type: FileType.image,
+              onFileLoading: (FilePickerStatus status) =>
+                  _logException(status.toString()),
+              allowedExtensions: ['jpg', 'png', 'gif'],
+              allowMultiple: true))
```
Xazin, line 104:
```suggestion
              allowMultiple: true,))
```
Comma for formatting, will make it easier to read.
Also maybe a good idea to split it up instead of using `()?.files` as it makes it a bit harder to read. _(just my opinion)_

lib/src/render/image/image_upload_widget.dart
```diff
@@ -84,8 +93,47 @@ class _ImageUploadMenuState extends State<ImageUploadMenu> {
...
+    } catch (e) {
+      _logException(e.toString());
+    }
+    if (!mounted) return;
+    setState(() {
+      _fileName = _paths != null ? _paths!.map((e) => e.name).toString() : null;
+
+      src = _paths != null ? _paths!.map((e) => e.path).toString() : null;
+      imageName = _fileName != null ? _fileName! : null;
+      widget.onSubmitted(src!);
```
Xazin, line 117: Adding eg. this class:
```
class FileInfo {
  String? name;
  String? src;
}
```

```suggestion
      if (paths != null) {
        final FileInfo fileInfo = _paths!.map((e) => FileInfo(fileName: e.name, src: e.path));
        
        imageName = fileInfo.name;
        
        if (fileInfo.src != null) {
          widget.onSubmitted(fileInfo.src);
        }
      }
```

lib/src/render/image/image_upload_widget.dart
```diff
@@ -84,8 +93,47 @@ class _ImageUploadMenuState extends State<ImageUploadMenu> {
...
+  }
+
+  void _logException(String msg) {
+    debugPrint(msg);
+  }
+
   @override
   Widget build(BuildContext context) {
+    TabController tabController =
+        TabController(initialIndex: 1, length: 2, vsync: this);
```
Xazin, line 136: Consider moving out of Build, so a rebuild doesn't cause it to change back to initial tab. 

You've commented out the dismissal, if intended this will cause a bug.

lib/src/render/image/image_upload_widget.dart
```diff
@@ -103,19 +151,45 @@ class _ImageUploadMenuState extends State<ImageUploadMenu> {
...
+            child: TabBarView(
+              controller: tabController,
+              children: [
+                Column(
+                  mainAxisAlignment: MainAxisAlignment.center,
+                  children: [
+                    const SizedBox(height: 15.0),
+                    _buildFileInput(context),
+                  ],
+                ),
```
Xazin, line 172: Consider abstracting into `FileImageInput`

lib/src/render/image/image_upload_widget.dart
```diff
@@ -103,19 +151,45 @@ class _ImageUploadMenuState extends State<ImageUploadMenu> {
...
+                Column(
+                  mainAxisAlignment: MainAxisAlignment.center,
+                  children: [
+                    _buildURLInput(),
+                    const SizedBox(height: 15.0),
+                    _buildUploadButton(context),
+                  ],
+                ),
```
Xazin, line 180: Consider abstracting into another widget as well

lib/src/render/image/image_upload_widget.dart
```diff
@@ -103,19 +151,45 @@ class _ImageUploadMenuState extends State<ImageUploadMenu> {
...
+          ),
           const SizedBox(height: 18.0),
-          _buildUploadButton(context),
         ],
       ),
     );
   }
 
-  Widget _buildHeader(BuildContext context) {
+  Widget _buildHeader(BuildContext context, String title) {
```
Xazin, line 190: I'm not sure what @LucasXu0 and the rest of the team thinks, but I'm actually not a big fan of methods returning widgets. 

If a Widget can have a `const` constructor, then it wouldn't rebuild on new builds, but if it's instead a method, then it will be reevaluated on every build no matter what.

In this case it can no longer be `const` due to `title`, but I thought it'd be nice to hear some opinions on this. I've seen it a lot in AppFlowy codebase.

lib/src/render/image/image_upload_widget.dart
```diff
@@ -24,11 +28,11 @@ void showImageUploadMenu(
           editorState: editorState,
           onSubmitted: (text) {
             // _dismissImageUploadMenu();
-            editorState.insertImageNode(text);
+            editorState.insertImageNode(text, 'file');
           },
           onUpload: (text) {
             // _dismissImageUploadMenu();
```
Xazin, line 34: Commented out dismissal of the popover, remove or use?

rileyhawk1417: Haven't really checked what _dismissImageUploadMenu just found it commented out

Xazin: Yeah fair enough, I guess we can remove it if it's not used. I assume it just dismisses the menu itself.

LucasXu0:
Make it as "Ready" once all things are set up.

lib/src/render/image/image_upload_widget.dart
LucasXu0: If the widget is reusable, I will create a Widget class instead of a Widget function. Also, depending on the complexity of the Widget, I agree with you that making it a const widget will improve performance.

PR #58, opened 2023-04-11
Allow developer to change toolbar color as well as option to show default toolbar items and html to document converter added

alihassan143:
this pr added option to add custom color to toolbar and elevation also exported extension so it can be accessed when adding custom toolbar items


LucasXu0:
Hey @alihassan143, please format the code before submitting, `dart format .`

lib/src/service/editor_service.dart
```diff
@@ -34,6 +34,9 @@ class AppFlowyEditor extends StatefulWidget {
     this.autoFocus = false,
     this.focusedSelection,
     this.customActionMenuBuilder,
+    this.toolbarColor=const Color(0xFF333333),
```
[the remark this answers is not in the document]
alihassan143: but inside toolbar widget color is already defined  if we did add color inside theme than its always be overridden by the already added color inside Material widget

alihassan143:
@LucasXu0  code is formated

alihassan143:
@LucasXu0  CompositedTransformFollower  instead of this widget if we use textselection toobar widget and pass childrens to its if the child list grows its adapts automatically to screen width

alihassan143:
@LucasXu0  toolbar elevation and toolbar color now moved to editor style

lib/appflowy_editor.dart
Xazin, line 1: Since you made a barrel file for `src/extenstions/...` then you can remove all duplicate exports.

eg. `export 'src/extensions/node_extensions.dart';` and `export 'src/extensions/attributes_extension.dart';`

alihassan143: done

LucasXu0:
This PR reminds me that I should split the toolbar function into a plugable feature. The editor and the toolbar should not have a reference to each other, but they can be composed together for use.

lib/appflowy_editor.dart
LucasXu0: There are still some pieces of code that need to be formatted. You can run dart fix --apply locally to apply the changes.

Xazin:
By the way, workflows have not run, but you need to reword all your commits to conform to the commit format.

It's a good practice to learn, I was happily surprised when I found out AppFlowy already enforces commit messages the way I prefer them.

Honestly I think for you it would be easier if you squashed all your commits into one, and then made the commit message eg. `feat: abstract toolbar colors to styling`.

If you want to do the squash, you can do it like this:
```
git reset --soft HEAD~12
git commit -m "feat: abstract toolbar colors to editor style"
```

If you want to reword each of your commits to conform to the commit lint, you can do:
```
git rebase HEAD~12
```
Then an interactive editor will open, press  `a` to start inserting, delete all the `pick` at the beginning and put an `r` instead. `r` is short for `reword`.

After. that hit ESC and then write `:wq!`, then **12 times** an editor will open, and you can press `a` to edit the commit message, then when done just do ESC and `:wq!`.

Thus preferring the squash..

Issue #56, opened 2023-04-11
[FR] Add the rules to analysis_options.yaml based on AppFlowy Flutter Style Guide


LucasXu0:
### Description

We should add rules to `analysis_options.ymal`, located in `frontend/analysis_options.yaml`, based on [AppFlowy Flutter Style Guide](https://appflowy.gitbook.io/docs/essential-documentation/contribute-to-appflowy/software-contributions/submitting-code/style-guides)

- [x] #57

### Impact

This will help format the code and reduce the time required for code reviewing.

#### Reference
* https://dart-lang.github.io/linter/lints/index.html
* https://appflowy.gitbook.io/docs/essential-documentation/contribute-to-appflowy/software-contributions/submitting-code/style-guides

**Please DO NOT close this issue**. We will keep this issue open to continue discussing which rules should be added.

Xazin:
I can work on this.

I have one question though, according to the style guide, we should declare the type for all variables. There is a lint for this: https://dart-lang.github.io/linter/lints/always_specify_types.html

I want to know your opinion on this specific lint before I add it, since it's quite a lot of work.

Xazin:
I'll do a PR for the rest of the lints once most of the PRs are merged, such that there is less work in rebasing.

PR #59, opened 2023-04-11
chore: add linters conforming to style guide

Xazin:
Relates: #56 #57


lib/src/flutter/overlay.dart
```diff
@@ -338,8 +352,12 @@ class OverlayState extends State<Overlay> with TickerProviderStateMixin {
...
   int _insertionIndex(OverlayEntry? below, OverlayEntry? above) {
     assert(above == null || below == null);
-    if (below != null) return _entries.indexOf(below);
-    if (above != null) return _entries.indexOf(above) + 1;
+    if (below != null) {
+      return _entries.indexOf(below);
+    }
+    if (above != null) {
+      return _entries.indexOf(above) + 1;
+    }
```
Xazin, line 360:
```suggestion
    if (below != null) {
      return _entries.indexOf(below);
    } else if (above != null) {
      return _entries.indexOf(above) + 1;
    }
```

lib/src/flutter/overlay.dart
```diff
@@ -430,25 +460,37 @@ class OverlayState extends State<Overlay> with TickerProviderStateMixin {
...
       'One or more of the specified entries are specified multiple times.',
     );
-    if (newEntriesList.isEmpty) return;
-    if (listEquals(_entries, newEntriesList)) return;
+    if (newEntriesList.isEmpty) {
+      return;
+    }
+    if (listEquals(_entries, newEntriesList)) {
+      return;
+    }
```
Xazin, line 493:
```suggestion
    if (newEntriesList.isEmpty || listEquals(_entries, newEntriesList)) {
      return;
    }
```

LucasXu0:
Hey, @Xazin. There are too many lints that have been added. I'm concerned that there will be conflicts with other PRs if I merge this. How about we add only one lint for this PR?

LucasXu0:
I believe that we should create a new issue to track each lint.

Xazin:
> I believe that we should create a new issue to track each lint.

How about these lints, all changes are from `require_trailing_commas`, and around 10-15 are from the others.

LucasXu0:
LGTM.

PR #60, opened 2023-04-11
fix: remove flutter overlay locally

Xazin:
We no longer need to have a local version of the Flutter Overlay in the project, as we are now running Flutter 3.7.x


LucasXu0:
LGTM. And I need to verify whether it works well on Flutter version 3.7.x as well

Issue #69, opened 2023-04-13
[BUG] Modify example application to fit on mobile


Xazin:
### Bug Description

In the example application, the theme applied to the Editor adds a lot of horizontal padding, which makes it hard and impossible to do any testing on Mobile.

### How to Reproduce

Run the example project on Android or iOS

### Expected Behavior

- [ ] SafeArea for top inset
- [ ] Less horizontal padding on Editor on Mobile (eg. 20 physical pixels)

### Operating System

Android, iOS

### AppFlowy Editor Version(s)

main

### Screenshots

<details>
  <summary>Screenshot from Android Emulator</summary>

![image](https://user-images.githubusercontent.com/42929161/231730368-5e463faf-1aad-4162-9fc5-cd78a56cb2dd.png)

</details>

Issue #68, opened 2023-04-13
[FR] Mobile Support


Xazin:
### Description

We will soon begin work related to supporting Mobile, therefore I have started mapping out some of the subtasks that relate. For some of the subtasks, it will be valid to open a separate issue due to the amount of work or complexity of the task.

The list of subtasks is subject to change, as work starts we will discover more items.

**Must haves:**
- [x] [BUG] Modify example application to fit on mobile
- [x] [BUG] Refactor InputService, KeyboardService, and Shortcuts, to accommodate mobile
- [x] [BUG] Popovers/Overlays such as the SelectionMenu is overflowing on mobile
- [ ] #307
- [x] [BUG] It is not possible to close the Soft Keyboard
- [x] [BUG] It is not possible to scroll in the editor on mobile
- [ ] [BUG] ScrollPosition should follow the cursor on mobile
- [ ] [BUG] Cannot delete a newline at the beginning of a delta (eg. TextDelta does not move up to the previous line, and cannot delete list/checkbox item)
- [x] [FR] Soft Keyboard should show a newline button instead of submit/done
- [x] [BUG] Add support for Query Schemes for Android and iOS in the example app, see: https://pub.dev/packages/url_launcher#configuration - Consider adding a section to the readme about this
- [x] #308

We should create separate issues for tasks that contain more complexity or is open to discussion.

### Impact

We will be able to support mobile not only with the editor itself but it is also required for https://github.com/AppFlowy-IO/appflowy to support mobile.

### Additional Context

This feature is part of the roadmap, we can create a milestone to track it, or add a new label eg. `mobile-support` or just `mobile` to make work related to this feature visible.

LucasXu0:
Thank you for your summary. Great job!

>  [BUG] Refactor InputService, KeyboardService, and Shortcuts, to accommodate mobile

FYI, I'm working on the second part.

rileyhawk1417:
Should we give the user an option for fullscreen mode?  Currently by default the editor hides the system notification bar.

allytiago:
It will also be very nice if there is optimized support for Samsung DeX. Desktop interface for Samsung devices
https://developer.samsung.com/samsung-dex

allytiago:
https://developer.samsung.com/samsung-dex

matthewwong525:
It seems like there's support mobile platform as of `1.0.0-dev0` . Could we get an update on this ticket? Is this ticket completed and are all the must haves covered?

[...]

PR #70, opened 2023-04-13
fix: lessen horizontal editor padding for mobile

Xazin:
Closes: #69 
Relates: #68


lib/src/render/style/editor_style.dart
```diff
@@ -179,7 +181,9 @@ class EditorStyle extends ThemeExtension<EditorStyle> {
   }
 
   static final light = EditorStyle(
-    padding: const EdgeInsets.fromLTRB(200.0, 0.0, 200.0, 0.0),
+    padding: Platform.isAndroid || Platform.isIOS
+        ? const EdgeInsets.symmetric(horizontal: 20)
+        : const EdgeInsets.fromLTRB(200.0, 0.0, 200.0, 0.0),
```
Xazin, line 186:
```suggestion
        : const EdgeInsets.symmetric(horizontal: 200),
```

Issue #72, opened 2023-04-13
[FR] Soft Keyboard should show a newline button instead of submit/done


Xazin:
### Bug Description

Currently, the soft keyboard (on-screen) shows a "go" action, this needs to be changed to `TextInputAction.newline`, so that it's possible to create newlines.

### How to Reproduce

1. Run the `example` project on Android or iOS
2. Click anywhere to open the soft keyboard
3. Try to make a newline...

### Expected Behavior

1. Run the `example` project on Android or iOS
2. Click anywhere to open the soft keyboard
3. Able to make a newline

### Operating System

Android, iOS

### AppFlowy Editor Version(s)

main

PR #73, opened 2023-04-13
feat: change soft keyboard action to newline

Xazin:
Closes: #72

Relates: #68


lib/src/service/input_service.dart
```diff
@@ -112,6 +112,7 @@ class _AppFlowyInputState extends State<AppFlowyInput>
           enableDeltaModel: true,
           inputType: TextInputType.multiline,
           textCapitalization: TextCapitalization.sentences,
+          inputAction: TextInputAction.newline,
```
LucasXu0, line 115: No rush to fix it now. I need some time to redesign the editor for the mobile platform a little bit.

Issue #76, opened 2023-04-14
[Bug] Image node render image first time on coming back to editor again start showing error again


alihassan143:
### Bug Description

when image added it shows only one time 
when came back from other screen to editor again 

### How to Reproduce

```
  editorState.insertImageNode(
                    "https://images.unsplash.com/photo-1471897488648-5eae4ac6686b?ixlib=rb-1.2.1&dl=sarah-dorweiler-QeVmJxZOv3k-unsplash.jpg&w=640&q=80&fm=jpg&crop=entropy&cs=tinysrgb");
```
    
when render that same document with image loaded again editor start showing error

### Expected Behavior

Editor need to render image again without error

### Operating System

iphone 14 pro simulator  

### AppFlowy Editor Version(s)

0.1.7

### Screenshots

https://user-images.githubusercontent.com/71745182/231955771-0e4d3eac-4553-4a4f-9b7d-622ff609c19b.mov

<img width="1680" alt="Screenshot 2023-04-14 at 11 10 14 AM" src="https://user-images.githubusercontent.com/71745182/231956036-bd63173f-928a-4c3e-ba1b-266b59824e66.png">

Xazin:
@LucasXu0 I can no longer reproduce on Android or iOS.

I believe we can close this.

Aruljebaraj:
any idea how to insert local images

PR #81, opened 2023-04-14
feat: abstract toolbar item icon sizes to editor style

alihassan143:
#76

### DIFF
--- a/lib/src/extensions/image_node_extension.dart
+++ b/lib/src/extensions/image_node_extension.dart
@@ -3,9 +3,10 @@ import 'package:appflowy_editor/appflowy_editor.dart';
 extension InsertImage on EditorState {
   void insertImageNode(String src) {
     final selection = service.selectionService.currentSelection.value;
-    if (selection == null) {
+    if (selection == null||selection.isCollapsed) {
       return;
     }
+ 
     final imageNode = Node(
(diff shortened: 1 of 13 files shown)


lib/src/extensions/image_node_extension.dart
```diff
@@ -0,0 +1,23 @@
+import 'package:appflowy_editor/appflowy_editor.dart';
+
+extension InsertImage on EditorState {
+  void insertImageNode(String src) {
+    final selection = service.selectionService.currentSelection.value;
```
LucasXu0, line 5: If the selection is not collapsed, it should also return.

alihassan143:
@LucasXu0 kindly check this pr

lib/src/render/image/image_upload_widget.dart
```diff
@@ -177,6 +177,9 @@ extension InsertImage on EditorState {
     if (selection == null) {
       return;
     }
+    if (selection.isCollapsed) {
```
Xazin, line 180:
```
if (selection == null || selection.isCollapsed) {
  return;
}
```
And weren't you supposed to add this check in the insert image handler extension?

Xazin (changes requested):

Issue #87, opened 2023-04-19
[FIX] Improve Ctrl+Backspace behaviour to delete word


MayurSMahajan:
### Description

 When Ctrl+Backspace is triggered in this sentence: 
`an apple   |` 
 the actual behavior is `an apple|` 
but the expected behavior should be: `an |`

### Impact

Every editor user, who wishes to delete words with ctrl+backspace on windows and linux and alt+backspace on macOS

PR #88, opened 2023-04-19
fix: improves cursor left word delete

MayurSMahajan:
Solves: #87 
This PR makes sure that by pressing ctrl+backspace in the editor, now also deletes the whitespace that it encounters before a word.

For given sentence: `an apple   |`
the actual behavior is `an apple|`
but the expected behavior should be: `an |`

This PR achieves the expected behavior.

Note: 
1. This PR moves the cursorLeftWordDelete and cursorLeftSentenceDelete handlers in a new file (previously stored in `arrow_keys_handler.dart` ).
2. This PR moves the extension methods written in `arrow_keys_handler.dart` file to a new file. 
3. Selection Range enum is now made public.


lib/src/extensions/position_extension.dart
```diff
@@ -0,0 +1,88 @@
+import 'package:appflowy_editor/appflowy_editor.dart';
+
+enum SelectionRange {
+  character,
+  word,
+}
+
+extension PositionExtension on Position {
+  Position? goLeft(
```
Xazin, line 9: I made goUp and goDown into `_moveVertical()`, you can include that in the extension as `moveVertical`, and then consolidate goUp and goDown into  one method as well.


The consolidated method could look something like this:
```
  Position? moveHorizontal(
    EditorState editorState, {
    bool moveLeft = true,
    _SelectionRange selectionRange = _SelectionRange.character,
  }) {
    final node = editorState.document.nodeAtPath(path);
    if (node == null) {
      return null;
    }

    if (moveLeft && offset == 0) {
      final previousEnd = node.previous?.selectable?.end();
      if (previousEnd != null) {
        return previousEnd;
      }

      return null;
    } else if (!moveLeft) {
      final end = node.selectable?.end();
      if (end != null && offset >= end.offset) {
        return node.next?.selectable?.start();
      }
    }

    switch (selectionRange) {
      case _SelectionRange.character:
        if (node is TextNode) {
          return Position(
            path: path,
            offset: moveLeft
                ? node.delta.prevRunePosition(offset)
                : node.delta.nextRunePosition(offset),
          );
        }

        return Position(path: path, offset: offset);
      case _SelectionRange.word:
        if (node is TextNode) {
          final result = moveLeft
              ? node.selectable?.getWordBoundaryInPosition(
                  Position(
                    path: path,
                    offset: node.delta.prevRunePosition(offset),
                  ),
                )
              : node.selectable?.getWordBoundaryInPosition(this);
          if (result != null) {
            return moveLeft ? result.start : result.end;
          }
        }

        return Position(path: path, offset: offset);
    }
  }
  ```

MayurSMahajan: okay

MayurSMahajan: do you mean, we should consolidate goLeft() and goRight() as well? 
If yes, then what would be the benefit of doing so, I think having two small separate methods are more readable than one big method

Xazin: I don't think readability is such a big issue for those two methods, it's 90% duplicate code.

MayurSMahajan: Yeah, I agree.

Xazin (approved):
Lgtm

Issue #89, opened 2023-04-20
[Bug] runtime error on web - unsupported operation, Platform._operatingSystem


stonymahony:
### Bug Description

I get an `Unsupported operation: Platform._operatingSystem` error for **web** with **Flutter 3.7.12** and **appflowy_editor 1.6, 1.8  and also 1.9**.

```
Thrown exception:
Unsupported operation: Platform._operatingSystem

Stack trace:
http://localhost:8080/C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart
266:49      throw_
http://localhost:8080/C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/_internal/js_dev_runtime/patch/io_patch.dart
244:5                   _operatingSystem
http://localhost:8080/C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/io/platform_impl.dart
56:40                                          get operatingSystem
http://localhost:8080/C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/io/platform.dart
64:45                                               get _operatingSystem
http://localhost:8080/C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart
872:8   get
http://localhost:8080/C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/io/platform.dart
149:51                                              get isAndroid
http://localhost:8080/C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart
872:8   get
http://localhost:8080/packages/appflowy_editor/src/render/style/editor_style.dart 194:33
get light
http://localhost:8080/C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart
872:8   get
http://localhost:8080/packages/appflowy_editor/src/render/style/editor_style.dart 6:19
get lightEditorStyleExtension
http://localhost:8080/packages/appflowy_editor/src/service/editor_service.dart 44:16
<fn>
http://localhost:8080/packages/appflowy_editor/src/service/editor_service.dart 45:16
new
```

### How to Reproduce

[...]

PR #91, opened 2023-04-22
fix: remove platform check in theme on web

Xazin:
Closes: #89


LucasXu0:
To prevent this issue from happening again, we should also run our tests on the web.

Issue #31, opened 2023-03-23
[FR] Support shift + click to update selection from start to end position


Xazin:
When you have cursor anywhere, hold shift and then click, you would expect for everything in between to be selected.

Currently it only moves cursor.

Xazin:
I made this https://github.com/Xazin/appflowy-editor/commit/e030733e11943c3ecad05f4ad8630aa5196731ab
But I'm not quite a fan of what I had to do to get it working.

Maybe someone else has ideas for improvements.

Xazin:
Updated my version on this commit https://github.com/AppFlowy-IO/appflowy-editor/commit/02c34904f84d656c93e8421ff9e4210f67942d80

PR #93, opened 2023-04-23
feat: shift+press to make selection

Xazin:
Relates: #31 

Opening the PR so we can discuss if it needs changes or if we need to consider alternative approaches.


Xazin:
Unfortunate spelling mistake 🤣

LucasXu0:
I will use another way to implement it.

PR #95, opened 2023-04-23
chore: prevent api misuse

a-wallen:
I noticed that we assert that both the `command` and `character` cannot both be set or both cannot be null. This PR hides that constructor and exposes two factory constructors, one with `command` and one with `character`.

Deletes `copyWith()`. To keep `copyWith` we might want to consider extending `ShortcutEvent` instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.1.9
+* feat: dark mode improvement by @hyj1204 in https://github.com/AppFlowy-IO/appflowy-editor/pull/90
+* feat: shift+tab to outdent a indented bulletpoint or checkbox. by @MayurSMahajan in https://github.com/AppFlowy-IO/appflowy-editor/pull/63
+* fix: enter to outdent checkbox/bullet lists by @Xazin in https://github.com/AppFlowy-IO/appflowy-editor/pull/84
+
 ## 0.1.8
 * chore: minor updates to documentation.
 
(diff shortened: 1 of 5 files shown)


a-wallen:
This change is likely to break AppFlowy. @LucasXu0, can this repo have "customer tests"? I.e, we run tests from AppFlowy or other projects that use appflowy-editor to see whether changes in the editor impact our customers?

LucasXu0:
@a-wallen Don't push your changes to the same commit in the main branch. 

* Why did you delete the `copyWith` function? 
* Can you provide more details on what you mean by "customer tests"?

a-wallen:
> @a-wallen Don't push your changes to the same commit in the main branch.
Yup, I think that I need to create a new PR with the changes.

> * Why did you delete the `copyWith` function?
`copyWith` only copies the command, not the character, if the command is null then we're just copying null? I can add it back.

> * Can you provide more details on what you mean by "customer tests"?
There's clients of the library. AppFlowy (the app) is an example. If we change APIs here, then AppFlowy will break if it's just using the latest version of this package. So, we should have tests that clone AppFlowy (customer) and try to run their integration tests against the new API from this package.

Xazin:
> > @a-wallen Don't push your changes to the same commit in the main branch.
> 
> Yup, I think that I need to create a new PR with the changes.
> 
> > * Why did you delete the `copyWith` function?
> 
> `copyWith` only copies the command, not the character, if the command is null then we're just copying null? I can add it back.
> 
> > * Can you provide more details on what you mean by "customer tests"?
> 
> There's clients of the library. AppFlowy (the app) is an example. If we change APIs here, then AppFlowy will break if it's just using the latest version of this package. So, we should have tests that clone AppFlowy (customer) and try to run their integration tests against the new API from this package.

Do we need this? AppFlowy uses a specific version of the editor and not the latest branch from git, therefore it isn't needed. It should be described in the changelog when there are breaking changes.

PR #99, opened 2023-04-24
feat: enter and option or alt to open link shortcut

MayurSMahajan:
Closes this issue [#2197](https://github.com/AppFlowy-IO/AppFlowy/issues/2197)

Description: 
Users can now navigate to a link and press Option/Alt + Enter to open the link in their browsers.

Note: 
Only works when the cursor selection is made with the keyboard since tapping the link with the mouse opens up Link Dialog.


lib/src/service/internal_key_event_handlers/link_open_handler.dart
```diff
@@ -0,0 +1,32 @@
...
+
+  if (linkText == null) {
+    return KeyEventResult.handled;
+  }
+
+  launchUrl(linkText);
+  return KeyEventResult.handled;
+};
+
+Future<void> launchUrl(String linkText) async {
```
LucasXu0, line 30: I suggest calling safeLaunchUrl directly. There is no special logic inside this function.

MayurSMahajan: The call to safeLaunchUrl directly needs to be `awaited`, which means we will have to make the handler function `async`, this results in a mismatch because the Shortcut event does not expect a method returning Future as a handler method, thus I separated the method into a new async method.

MayurSMahajan: okay

PR #100, opened 2023-04-24
fix: do not trigger link menu when selection is collapsed

Xazin:
Closes: https://github.com/AppFlowy-IO/AppFlowy/issues/2233


Xazin:
Merely codecov step that fails, not actual tests

Xazin:
I guess in the future, it would be better to have the link menu ask for text, and by default just have the text of what was selected, if nothing is selected it should just be empty.

<details>
  <summary>Confluence with and without Selection</summary>

image

image

</details>

Xazin:
Rebased and added an expect in the test. Still ready to be merged.

PR #102, opened 2023-04-25
fix: use target node attributes on paste

Xazin:
Closes: https://github.com/AppFlowy-IO/AppFlowy/issues/2276


LucasXu0:
@Xazin Could you help to add some test for it?

Xazin:
> @Xazin Could you help to add some test for it?

Will do

Issue #65, opened 2023-04-12
[FR] Support local image picker


rileyhawk1417:
### Description

Pick images from local computer, and use them in AppFlowy-Editor.

### Impact

Users are able to add images from their device

LucasXu0:
closed as completed. 

<img width="619" alt="Screenshot 2023-07-12 at 17 03 16" src="https://github.com/AppFlowy-IO/appflowy-editor/assets/11863087/474a1ef1-a146-4c70-87fa-e7d6b5d40897">

PR #111, opened 2023-05-05
Local image plugin

rileyhawk1417:
Added local image support for AppFlowy Editor, so you can place images from  your computer into the editor.
Also separated the tests into two files one for local images & the other for network images.
This is meant to solve issue #65


test/render/image/image_upload_widget_test.dart
```diff
@@ -33,7 +33,7 @@ void main() {
         Selection.single(path: [0], startOffset: 19),
       );
 
-      editor.editorState.insertImageNode('no_src');
+      editor.editorState.insertImageNode('no_src', 'no_type');
```
rileyhawk1417, line 36: Not sure why we need to check for NULL values here. Although there is a widget that notifies the user if image placement has failed.

lib/src/render/image/image_node_widget.dart
```diff
@@ -138,30 +141,50 @@ class ImageNodeWidgetState extends State<ImageNodeWidget> with SelectableMixin {
...
-      widget.src,
-      width: _imageWidth == null ? null : _imageWidth! - _distance,
-      gaplessPlayback: true,
-      loadingBuilder: (context, child, loadingProgress) {
-        if (loadingProgress == null ||
-            loadingProgress.cumulativeBytesLoaded ==
-                loadingProgress.expectedTotalBytes) {
-          return child;
+    Image? image;
+
```
Xazin, line 145:
```suggestion
    Image? image;
    final width = _imageWidth == null ? null : _imageWidth! - _distance; 

```
You can move this ternary operation outside of the switch so it's the same for both cases, you don't have to though.

lib/src/render/image/image_node_widget.dart
```diff
@@ -138,30 +141,50 @@ class ImageNodeWidgetState extends State<ImageNodeWidget> with SelectableMixin {
...
+        if (!imageFile.existsSync()) {
+          return const SizedBox.shrink();
         }
+        image = Image.file(
+          imageFile,
+          width: _imageWidth == null ? null : _imageWidth! - _distance,
+          gaplessPlayback: true,
+          errorBuilder: (context, error, stackTrace) {
+            return _buildError(context);
+          },
```
Xazin, line 158:
```suggestion
          errorBuilder: (context, _, __) => _buildError(context),
```
Expression body

lib/src/render/image/image_upload_widget.dart
```diff
@@ -78,8 +91,50 @@ class _ImageUploadMenuState extends State<ImageUploadMenu> {
...
+      _paths = (await FilePicker.platform.pickFiles(
+        type: FileType.image,
+        onFileLoading: (FilePickerStatus status) =>
+            _logException(status.toString()),
+        allowedExtensions: ['jpg', 'png', 'gif'],
+        allowMultiple: true,
+      ))
+          ?.files;
+    } on PlatformException catch (e) {
+      _logException('Unsupported Operation  ${e.toString()}');
```
Xazin, line 106: I'd prefer just having `debugPrint` since all `_logException` is debugPrint, and it's not clear what is happening

lib/src/render/image/image_upload_widget.dart
```diff
@@ -78,8 +91,50 @@ class _ImageUploadMenuState extends State<ImageUploadMenu> {
...
+    _resetState();
+    try {
+      _paths = (await FilePicker.platform.pickFiles(
+        type: FileType.image,
+        onFileLoading: (FilePickerStatus status) =>
+            _logException(status.toString()),
+        allowedExtensions: ['jpg', 'png', 'gif'],
+        allowMultiple: true,
+      ))
+          ?.files;
```
Xazin, line 104: 1. Why allow multiple, when you only take the `last` image in the next action?

2. I'm wondering if it wouldn't be better to have a BLoC that has an event `ImageUploadEvent.filePicked(file: imageFile)` that takes the result of `final imageFile = await FilePicker.platform.pickFiles()?.files.first` - and only if there is an imageFile of course.

lib/src/render/image/image_upload_widget.dart
```diff
@@ -148,6 +228,25 @@ class _ImageUploadMenuState extends State<ImageUploadMenu> {
...
+      height: 48,
+      child: TextButton(
+        style: ButtonStyle(
+            backgroundColor: MaterialStateProperty.all(const Color(0xFF00BCF0)),
+            shape: MaterialStateProperty.all<RoundedRectangleBorder>(
+                RoundedRectangleBorder(
+                    borderRadius: BorderRadius.circular(12.0)))),
+        onPressed: _pickFiles,
+        child: const Text(
+          'Pick from computer',
```
Xazin, line 243: Localization

lib/src/render/image/image_upload_widget.dart
rileyhawk1417: I don't think `final imageFile = await FilePicker.platform.pickFiles()?.files.first` will work. Since its an await function but it would could work like this: `final imageFile = (await FilePicker.platform.pickFiles())?.files.first`

Issue #112, opened 2023-05-05
[Bug] Indented entries are missing from the exported markdown


Amartya-Mahato:
### Bug Description

Referring to [this](https://github.com/AppFlowy-IO/AppFlowy/issues/2452#issue-1694041367) issue.
Not being able to parse indented entries.
![image](https://user-images.githubusercontent.com/72689903/236567704-e6508460-49d7-4d63-aa3c-39e33bd3da94.png)
The output is
![image](https://user-images.githubusercontent.com/72689903/236567781-ea72a857-c590-4b10-9acb-fe159d310ad0.png)

### How to Reproduce

Indenting entries causing this issue

- [ ] one entry 
  - [ ] second entry
- [ ] last entry

----or----

* one entry
  * second entry
* last entry

### Expected Behavior

All entries should be recorded with the necessary indentation.

### Operating System

Windows 11

### AppFlowy Editor Version(s)

appflowy_editor 0.1.9

PR #113, opened 2023-05-06
Fixes markdown - indentation/level issue

saru2020:
Fixes https://github.com/AppFlowy-IO/appflowy-editor/issues/112

AppFlowy_FixesMarkdownIndentationLevelIssue

### DIFF
--- a/lib/src/plugins/markdown/encoder/parser/text_node_parser.dart
+++ b/lib/src/plugins/markdown/encoder/parser/text_node_parser.dart
@@ -17,7 +17,6 @@ class TextNodeParser extends NodeParser {
     final attributes = textNode.attributes;
     var result = markdown;
     var suffix = '\n';
-
     if (attributes.isNotEmpty &&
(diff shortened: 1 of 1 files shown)


Xazin:
I think there is a bit of a cleaner way to implement this, that allows for all levels of nesting, whereas this solution only support level 1.

Would it be okay if I open a PR to replace this Saru?

saru2020:
> I think there is a bit of a cleaner way to implement this, that allows for all levels of nesting, whereas this solution only support level 1.
> 
the solution supports all levels.

> Would it be okay if I open a PR to replace this Saru?
> 
> See: [Xazin@991e28b](https://github.com/Xazin/appflowy-editor/commit/991e28bd69a6b67a955f1c868c5f82c231ee753e)

but, your solution looks neat + you've added some tests as well, please go ahead with that, thanks!

Xazin:
Opened #115

PR #115, opened 2023-05-06
fix: nested text node parser

Xazin:
Closes: #112

Replaces: #113


LucasXu0:
LGTM. @Xazin Can you help to implement this logic in mobile branch too?

PR #119, opened 2023-05-14
fix: fixed image placeholder border color

nurmukhametdaniyar:
The image placeholder color now matches Divider's color, so it will be visible both on the light and dark themes

image

image


nurmukhametdaniyar:
Fixes https://github.com/AppFlowy-IO/AppFlowy/issues/2362

PR #125, opened 2023-05-18
fix: allow multiple characters in shortcuts

Xazin:
This cuts down on code duplication by allowing multiple characters to trigger one shortcut event.

**Note:**
We can discuss how we should split the characters, currently I split the character string by `,` and remove all spaces. This removes ` ` as a possibility to be included in a multi character command. We can just as easily build a RegExp pattern to do the matching if we want. (Or simply just have each chararacter be evaluated individually like `abc` would be `a`, `b`, and `c`.


lib/src/service/keyboard_service.dart
```diff
@@ -158,7 +156,18 @@ class _AppFlowyKeyboardState extends State<AppFlowyKeyboard>
...
+      }
+
+      return event.character == character;
+    }
+
+    return keybindings.containsKeyEvent(event);
   }
+
+  List<String> _charactersFromString(String character) =>
+      character.split(',').map((c) => c.replaceAll(' ', '')).toList();
```
LucasXu0, line 172:
```dart
bool canRespondToRawKeyEvent(RawKeyEvent event) {
    final character = this.character;
    if (character != null && character.isNotEmpty) {
      final characters = character.split(',').map((e) => e.trim());
      return characters.contains(event.character);
    }
    return keybindings.containsKeyEvent(event);
  }
```

LucasXu0: I think it's more clear.

LucasXu0:
LGTM.

Xazin:
@LucasXu0 I almost forgot about this, it's ready.

LucasXu0:
would you mind re-opening a new PR to refactor the `CharacterShortcutEvent`?

Xazin:
> would you mind re-opening a new PR to refactor the `CharacterShortcutEvent`?

I'll refactor it accordingly sometime soon. :+1:

MichalNemec:
Hello, any news regarding this? i would like to have custom character shortcut for #whatever

PR #126, opened 2023-05-19
fix: overflow error in mobile

hyj1204:
Reduce the padding in the editor style on mobile and set the mobile entry point in Editor

image


hyj1204:
close PR to update merge  target branch

PR #137, opened 2023-05-27
fix: new editor bugs

LucasXu0:


lib/src/editor/editor_component/service/renderer/block_component_container.dart
```diff
@@ -65,31 +65,6 @@ class BlockComponentContainerState extends State<BlockComponentContainer>
       ),
     );
 
-    // if (widget.showBlockComponentActions) {
```
Xazin, line 68: Can we delete this?

Xazin (approved):

lib/src/editor/editor_component/service/renderer/block_component_container.dart
LucasXu0: sure.

LucasXu0: This logic has moved into each block component. So I just deleted it.

Issue #162, opened 2023-06-03
[Bug] In tollbar instead of Embed code it is displaying Strikethrough


Mukund-Tandon:
### Bug Description

When we select text and hover over the embed code button it shows Strikethrough instead of Embed code

### How to Reproduce

Select text and hover over embed code button

### Expected Behavior

Should show Embed code 

### Operating System

Linux (Fedora Workstation 35)

### AppFlowy Editor Version(s)

1.0.0-dev.1

### Screenshots

![Screenshot from 2023-06-03 20-39-40](https://github.com/AppFlowy-IO/appflowy-editor/assets/71614009/cd1f9236-aab0-4410-aed8-2fe69f9e5c31)

PR #163, opened 2023-06-03
fix: fixed toolbar item tooltip desciption for embed code button

Mukund-Tandon:
fixes - #162 
Before-

Screenshot from 2023-06-03 20-39-40

After-

Screenshot from 2023-06-03 20-56-26


LucasXu0:
LGTM.

PR #191, opened 2023-06-12
feat: rtl and auto direction

zoli:
part of https://github.com/AppFlowy-IO/AppFlowy/issues/39

https://github.com/AppFlowy-IO/appflowy-editor/assets/3286303/de720145-7c0b-4940-9692-a23510a884c0

I had to set fill the available width of every `FlowyRichtText` parent container so the text could start from the most right of the window.

Introducing RTL and Auto direction for each node. In Auto direction, it checks for the first character with strong directionality.


LucasXu0:
Nice try!

lib/src/editor/block_component/bulleted_list_block_component/bulleted_list_block_component.dart
```diff
@@ -87,10 +87,14 @@ class _BulletedListBlockComponentWidgetState
   Widget buildComponent(BuildContext context) {
     Widget child = Container(
       color: backgroundColor,
+      width: double.infinity,
```
LucasXu0, line 90: Why using `double.infinity` here?

LucasXu0:
I suggest adding two ways to enable the RTL (right-to-left) feature. One is the way you mentioned, and the other one is by exposing a new parameter in the editor called "enableRtl". If the developer enables this parameter, then all the text inside the editor should be displayed in RTL format.

lib/src/editor/block_component/bulleted_list_block_component/bulleted_list_block_component.dart
zoli: Because it won't go to the right side of the page without it. It will become RTL on the left side of the page.

zoli:
> I suggest adding two ways to enable the RTL (right-to-left) feature. One is the way you mentioned, and the other one is by exposing a new parameter in the editor called "enableRtl". If the developer enables this parameter, then all the text inside the editor should be displayed in RTL format.

That's a good idea, Also I was thinking other than the editor scope base setting we could have some sort of page scope setting and set only the whole specific page direction to RTL.

So it will be like this:

1. Editor scope
2. Page scope
3. Per line scope

LucasXu0:
What's the differences between editor scope and page scope? actually, the last one should block scope ( same as node scope, like paragraph, heading an so on )

zoli:
> What's the differences between editor scope and page scope? 

When the direction is set to RTL in the editor scope every page will get viewed RTL by default (the direction will get default to RTL). But when the direction is set to RTL in the page scope every block on that page will get viewed as RTL by default. The same would be if the direction is set to AUTO or LTR in the editor/page scope.

> actually, the last one should block scope ( same as node scope, like paragraph, heading an so on )

Yes, You are right.

zoli:
Made this draft because I saw selection movement doesn't work correct on rtl sections.

zoli:
This is ready. Also, I updated the showcase video.
Now we have Auto, RTL, LTR direction.

/cc @LucasXu0

LucasXu0:
Hey, @zoli. I refactored some parts of the code.

1. delete the appflowy_rich_text_key_dir. Use `blockComponentTextDirection` instead.
```dart
/// the text direction of the block component
///
/// its value must be one of the following:
///   - [blockComponentTextDirectionLTR] or 'ltr': left to right
///   - [blockComponentTextDirectionRTL] or 'rtl': right to left
///   - [blockComponentTextDirectionAuto] or auto: depends on the text
///
/// only works for the block with text,
///   e.g. paragraph, heading, quote, to-do list, bulleted list, numbered list
const blockComponentTextDirection = 'textDirection';
const blockComponentTextDirectionAuto = 'auto';
const blockComponentTextDirectionLTR = 'ltr';
const blockComponentTextDirectionRTL = 'rtl';
```

2. refactored the get text direction function. Please check it.
3. move the level calculation to node extension.
4. refactored the padding block component, use stateless widget instead.

lib/src/editor/block_component/base_component/text_direction_mixin.dart
```diff
@@ -0,0 +1,69 @@
...
+    // if the block component doesn't has a text direction attribute, but has a
+    //  parent, use the text direction of the parent
+    final previousNodeContainsTextDirection = node.previousNodeWhere(
+      (element) => element.attributes.containsKey(blockComponentTextDirection),
+    );
+    if (value == blockComponentTextDirectionAuto &&
+        previousNodeContainsTextDirection != null) {
+      final previousValue = previousNodeContainsTextDirection
+          .attributes[blockComponentTextDirection];
+      return previousValue.toTextDirection(fallback: defaultTextDirection);
```
zoli, line 33: @LucasXu0 It should change the default direction not return it.
Also, this currently throws an exception that "Class 'String' has no instance method 'toTextDirection'".

This is a test case that shows the problem:
```dart
test('auto empty text use previous node direction', () {
      final node = pageNode(
        children: [
          paragraphNode(
            text: 'Before node',
            textDirection: blockComponentTextDirectionRTL,
          ),
          paragraphNode(
            text: '',
            textDirection: blockComponentTextDirectionAuto,
          ),
        ],
      );
      final direction =
          TextDirectionTest(node: node.children.last).calculateTextDirection(
        defaultTextDirection: TextDirection.ltr,
      );
      expect(direction, TextDirection.rtl);
    });
```

zoli:
Hey, @LucasXu0. I haven't checked your changes yet (will do it tonight).
Should I commit my changes or do you prefer to mention them as a comment/review here on the PR?

lib/src/editor/block_component/base_component/text_direction_mixin.dart
zoli: This applies not only to empty text cases but also to cases where the text has no strong directionality (e.g. characters like '$', '#', etc).

LucasXu0:
Please mention them as a review comment and we can discuss about it.

lib/src/editor/block_component/base_component/text_direction_mixin.dart
LucasXu0: Sounds reasonable. Please help me to correct it.

LucasXu0: I thought "Auto" meant that the text direction should be based on the previous one and itself. Is that correct?

zoli: Auto means determining the direction based on itself (the text) if it's not determinable it should fallback to 
1. what the direction of the line was before (going from "$ Appflowy" to "$" the direction before is LTR)
2. previous node
3. default direction

And this is just made up by me from what I think makes sense. I haven't seen any documented convention.

lib/src/editor/block_component/text_block_component/text_block_component.dart
```diff
@@ -91,21 +97,39 @@ class _TextBlockComponentWidgetState extends State<TextBlockComponentWidget>
   @override
   Node get node => widget.node;
 
+  String? lastStartText;
+  TextDirection? lastDirection;
+
   @override
   Widget buildComponent(BuildContext context) {
+    final textDirection = calculateTextDirection(
+      defaultTextDirection: Directionality.of(context),
```
zoli, line 106: Isn't better to use `maybeOf`? For the case that "there is no [Directionality] ancestor widget" it throws an exception (from the documentation).

lib/src/editor/block_component/base_component/text_direction_mixin.dart
zoli: > Sounds reasonable. Please help me to correct it.

For the exception 
```dart 
final previousValue = previousNodeContainsTextDirection.attributes[blockComponentTextDirection];
``` 
should have `String` type like this 
```dart
final String previousValue = previousNodeContainsTextDirection.attributes[blockComponentTextDirection];
```

Also, my test wasn't complete. It was just showing the exception problem the more complete test which shows the direction calculation problem is this:
```dart
test('auto empty text use previous node direction', () {
      final node = pageNode(
        children: [
          paragraphNode(
            text: 'سلام',
            textDirection: blockComponentTextDirectionRTL,
          ),
          paragraphNode(
            text: '\$',
            textDirection: blockComponentTextDirectionAuto,
          ),
        ],
      );
      final direction =
          TextDirectionTest(node: node.children.last).calculateTextDirection(
        defaultTextDirection: TextDirection.ltr,
      );
      expect(direction, TextDirection.rtl);
    });

test('auto empty text don\'t use previous node direction because we can determine by the node text', () {
      final node = pageNode(
        children: [
          paragraphNode(
            text: 'سلام',
            textDirection: blockComponentTextDirectionRTL,
          ),
          paragraphNode(
            text: 'Hello',
            textDirection: blockComponentTextDirectionAuto,
          ),
        ],
      );
      final direction =
          TextDirectionTest(node: node.children.last).calculateTextDirection(
        defaultTextDirection: TextDirection.rtl,
      );
      expect(direction, TextDirection.ltr);
    });
```

I suggest adding these two tests to "text_direction_mixin_test". For correcting it I suggest changing 
```dart
return previousValue.toTextDirection(fallback: defaultTextDirection);
```
to
```dart 
defaultTextDirection =
          previousValue.toTextDirection(fallback: defaultTextDirection);
```

lib/src/editor/block_component/base_component/text_direction_mixin.dart
```diff
@@ -0,0 +1,69 @@
...
+  r"(?:([\p{sc=Arabic}\p{sc=Hebrew}\p{sc=Syriac}\p{sc=Thaana}])|([0-9\p{sc=Armenian}\p{sc=Bengali}\p{sc=Bopomofo}\p{sc=Braille}\p{sc=Buhid}\p{sc=Canadian_Aboriginal}\p{sc=Cherokee}\p{sc=Cyrillic}\p{sc=Devanagari}\p{sc=Ethiopic}\p{sc=Georgian}\p{sc=Greek}\p{sc=Gujarati}\p{sc=Gurmukhi}\p{sc=Han}\p{sc=Hangul}\p{sc=Hanunoo}\p{sc=Hiragana}\p{sc=Inherited}\p{sc=Kannada}\p{sc=Katakana}\p{sc=Khmer}\p{sc=Lao}\p{sc=Latin}\p{sc=Limbu}\p{sc=Malayalam}\p{sc=Mongolian}\p{sc=Myanmar}\p{sc=Ogham}\p{sc=Oriya}\p{sc=Runic}\p{sc=Sinhala}\p{sc=Tagalog}\p{sc=Tagbanwa}\p{sc=Tamil}\p{sc=Telugu}\p{sc=Thai}\p{sc=Tibetan}\p{sc=Yi}]))",
+  unicode: true,
+);
+
+mixin BlockComponentTextDirectionMixin {
+  Node get node;
+
+  /// Calculate the text direction of a block component.
+  TextDirection calculateTextDirection({
+    TextDirection defaultTextDirection = TextDirection.ltr,
```
zoli, line 16: We don't have lastDirection (`Directionality.off` doesn't give it) so we cant have the first item in fallback order I mentioned: 

1. what the direction of the line was before
2. previous node
3. default direction

So think you start writing RTL text in block component with AUTO direction. After first character you press backspace and clear the whole text. The direction will suddenly change from RTL to LTR.

This is the test I wrote for this case and failed
```dart
testWidgets("clear text but keep the old direction", (tester) async {
      final editor = tester.editor
        ..addNode(
          paragraphNode(
            text: 'Hello',
            textDirection: blockComponentTextDirectionLTR,
          ),
        )
        ..addNode(
          paragraphNode(
            text: 'س',
            textDirection: blockComponentTextDirectionAuto,
          ),
        );
      await editor.startTesting();

      Node node = editor.nodeAtPath([1])!;
      expect(
        node.selectable?.textDirection().name,
        blockComponentTextDirectionRTL,
      );

      final selection = Selection.collapsed(
        Position(path: [1], offset: 1),
      );
      await editor.updateSelection(selection);

      await simulateKeyDownEvent(LogicalKeyboardKey.backspace);
      await tester.pumpAndSettle();

      node = editor.nodeAtPath([1])!;
      expect(
        node.delta?.toPlainText().isEmpty,
        true,
      );
      expect(
        node.selectable?.textDirection().name,
        blockComponentTextDirectionRTL,
      );

      await editor.dispose();
    });
```

lib/src/editor/block_component/base_component/text_direction_mixin.dart
```diff
@@ -0,0 +1,69 @@
+import 'package:appflowy_editor/appflowy_editor.dart';
+import 'package:flutter/material.dart';
+
+/// This regular expression matches either a single character from specific RTL (right-to-left) script categories or a digit or character from various other script categories.
+/// It can be useful for identifying or manipulating text with mixed directional properties.
+final _regex = RegExp(
```
zoli, line 6: Out of curiosity, I ran a simple benchmark with some simplifications.
The regex is costly even when defining it as a final global variable.

The average run for matching regex each time on my system was 2.6500 ms but without the matching each time (using last start text for example) it was 910 μs.

Although I don't know how much this affects user experience.

This was my benchmark code:
```dart
import 'package:benchmarking/benchmarking.dart';

final _regex = RegExp(
  r"(?:([\p{sc=Arabic}\p{sc=Hebrew}\p{sc=Syriac}\p{sc=Thaana}])|([0-9\p{sc=Armenian}\p{sc=Bengali}\p{sc=Bopomofo}\p{sc=Braille}\p{sc=Buhid}\p{sc=Canadian_Aboriginal}\p{sc=Cherokee}\p{sc=Cyrillic}\p{sc=Devanagari}\p{sc=Ethiopic}\p{sc=Georgian}\p{sc=Greek}\p{sc=Gujarati}\p{sc=Gurmukhi}\p{sc=Han}\p{sc=Hangul}\p{sc=Hanunoo}\p{sc=Hiragana}\p{sc=Inherited}\p{sc=Kannada}\p{sc=Katakana}\p{sc=Khmer}\p{sc=Lao}\p{sc=Latin}\p{sc=Limbu}\p{sc=Malayalam}\p{sc=Mongolian}\p{sc=Myanmar}\p{sc=Ogham}\p{sc=Oriya}\p{sc=Runic}\p{sc=Sinhala}\p{sc=Tagalog}\p{sc=Tagbanwa}\p{sc=Tamil}\p{sc=Telugu}\p{sc=Thai}\p{sc=Tibetan}\p{sc=Yi}]))",
  unicode: true,
);

void main() {
  final numbers = List.generate(10000, (i) => i);
  final randomizedKeys = numbers.toList()..shuffle();
  String? result; // so that the loop isn't optimized out

  // This is a "functional" benchmark definition.
  // Alternatively, you can define your benchmark as a class, by overriding either SyncBenchmark or AsyncBenchmark.
  syncBenchmark(
    'Without last start text',
    () => randomizedKeys.forEach((i) => result = determineDirection("a$i")),
  ).report(units: numbers.length);
  syncBenchmark(
    'With last start text',
    () => randomizedKeys.forEach(
      (i) => result =
          determineDirection("a$i", lastStartText: 'a', lastDirection: 'rtl'),
    ),
  ).report(units: numbers.length);
}

String determineDirection(String text,
    {String? lastStartText, String? lastDirection}) {
  if (lastStartText != null &&
      lastDirection != null &&
      text.startsWith(lastStartText)) {
    return lastDirection;
  }

  final matches = _regex.firstMatch(text);
  if (matches != null) {
    if (matches.group(1) != null) {
      return 'rtl';
    } else if (matches.group(2) != null) {
      return 'ltr';
    }
  }

  return 'ltr';
}
```

zoli:
I don't have any comments on other changes, just on `calculateTextDirection` function which I added above. If you agree with me about the fallback order there are some more modifications to preserve it.

lib/src/editor/block_component/text_block_component/text_block_component.dart
LucasXu0: Sure.

LucasXu0:
> I don't have any comments on other changes, just on `calculateTextDirection` function which I added above. If you agree with me about the fallback order there are some more modifications to preserve it.

Yep. I agreed that. You can just modify them.

1. add the tests you mentioned
2. optimize the `calculateTextDirection` function

lib/src/editor/block_component/base_component/text_direction_mixin.dart
LucasXu0: The results on my side are very close.

We can optimize it later because I don't think using `lastStartText` and `lastDirection` is a good choice.


```
Without last start text
          total runs:      2 369
          total time:     2.0008  s
         average run:        844 μs
         runs/second:    1 184.8
               units:     10 000
        units/second: 11 848 341
       time per unit:     0.0844 μs

With last start text
          total runs:      5 947
          total time:     2.0001  s
         average run:        336 μs
         runs/second:    2 976.2
               units:     10 000
        units/second: 29 761 905
       time per unit:     0.0336 μs
```

Issue #170, opened 2023-06-05
[FR] refactory image block component 


rileyhawk1417:
### Description

To refactor the image block component. By doing splitting up the large component trees into smaller widgets.

### Impact

It will be easier to maintain, as separate components will have one job & one job only to do.

PR #195, opened 2023-06-14
Refactor image block

rileyhawk1417:
This PR is related to #170 
Moving the image block function helpers into another file, and placing them into classes.


LucasXu0:
@rileyhawk1417 there're some analysis errors.

rileyhawk1417:
Hi @Xazin there is a [PR](https://github.com/AppFlowy-IO/appflowy-editor/pull/232) that has several breaking changes. Am trying to refactor the code now to support those changes. Have figured it out the solution just need to write it. Then squash the commits.

rileyhawk1417:
@Xazin & @LucasXu0 I have moved this PR into another one that has all the commits squashed.

Issue #202, opened 2023-06-15
[Bug] Unable to turn highlighted text to no highlight, and colored text to default text easily.


liyaowhen:
### Bug Description

When you highlight or color text, there is no easy way to revert your changes without using Ctrl-Z or using the custom color option.

### How to Reproduce

Highlight or color text with a different color. You will see that it is hard to change the text back to its original non-formatted state without using Ctrl-Z.

### Expected Behavior

There should be an option to restore the default text (non-highlighted text or regular black text)

### Operating System

Windows 11

### AppFlowy Editor Version(s)

0.2.2

### Screenshots

![AppFlowy_4y24d7qeXV](https://github.com/AppFlowy-IO/appflowy-editor/assets/72400519/9049c4ba-94b8-42ed-af0f-c4818cd0daf3)

### Additional Context

i'm going to fix it myself

annieappflowy:
I've seen a reset to default option in the past...
@LucasXu0

the-jasoney:
> I've seen a reset to default option in the past... @LucasXu0

It seems to have been deleted in commit 8488c86

PR #203, opened 2023-06-15
fix: #202

liyaowhen:
- added option to set the highlight color of text to "Default", making the highlight white
- added option to set the text color of highlighted text to "Default", making the text black


liyaowhen:
the cla agreement is not updating 😭

liyaowhen:
I was using my old email, so it did not recognize me as a contributor

liyaowhen:
see #204

PR #204, opened 2023-06-16
fix: #202

liyaowhen:
adds ability to remove text highlight and color by adding a default option for both menus


liyaowhen:
i closed the #203 because I used the wrong account :(

LucasXu0:
The default background color should be `Colors.transparent` instead of `Colors.white`.

LucasXu0:
And the default text color should be same as the text color from `TextStyleConfiguration`

liyaowhen:
ok, ill make another pull request when I fix it :)

liyaowhen:
oh, i just realized i can add more to the pull request :)

liyaowhen:
nevernmind, my build setup is broken.

PR #208, opened 2023-06-16
fix: added default options to both text color and background color me…

the-jasoney:
Adds a default option to both text color and text background color menus.


the-jasoney:
Fixes issues raised in #204

LucasXu0:
@the-jasoney Please format the code.

PR #213, opened 2023-06-17
fix: behaviour of home and end keys on windows

MayurSMahajan:
## Default Behavior of Home and End Key on Windows/Linux

The default behavior of the home key on Windows/Linux is to navigate the cursor to the start of a line.
The default behavior of the home key on MacOS is to scroll the page to the top of the document.
Currently, we only support the MacOS behavior, this PR aims to make the cursor navigate to the start of a line on Windows/Linux.

It also introduces a new shortcut `Ctrl+Home` to scroll to the top of the document in Windows and `Ctrl+End` to scroll to the bottom of the document. 

The default behavior for MacOS is retained:
 - `Ctrl+arrow keys` to go to the beginning or end of a line.
 - `Home` and `End`keys to scroll in the document.

Solves [#2817](https://github.com/AppFlowy-IO/AppFlowy/issues/2817).


LucasXu0:
LGTM

PR #214, opened 2023-06-19
fix: toolbar position at most top

zoli:
At most top toolbar position goes out of the screen. Added a condition to check and position it on the bottom of the line.

https://github.com/AppFlowy-IO/appflowy-editor/assets/3286303/bf8b4d44-f897-4731-bd62-9343a1a16917

https://github.com/AppFlowy-IO/appflowy-editor/assets/3286303/8ae6fbec-1f86-4cfa-8c2f-ead14a4ae11b

### DIFF
--- a/lib/src/editor/toolbar/desktop/floating_toolbar.dart
+++ b/lib/src/editor/toolbar/desktop/floating_toolbar.dart
@@ -149,9 +149,7 @@ class _FloatingToolbarState extends State<FloatingToolbar>
     }
 
     final rect = _findSuitableRect(rects);
-    print(rect);
     final (top, left, right) = calculateToolbarOffset(rect);
-    print((top, left, right));
     _toolbarContainer = OverlayEntry(
       builder: (context) {
         return Positioned(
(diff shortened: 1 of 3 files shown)


lib/src/editor/toolbar/desktop/floating_toolbar.dart
```diff
@@ -149,7 +149,9 @@ class _FloatingToolbarState extends State<FloatingToolbar>
     }
 
     final rect = _findSuitableRect(rects);
+    print(rect);
```
LucasXu0, line 152: please remove the print function

lib/src/editor/toolbar/desktop/floating_toolbar.dart
```diff
@@ -149,7 +149,9 @@ class _FloatingToolbarState extends State<FloatingToolbar>
...
     final (top, left, right) = calculateToolbarOffset(rect);
+    print((top, left, right));
```
LucasXu0, line 154: same here

zoli: Thanks for fixing them.

PR #220, opened 2023-06-20
fix: mobile selection scroll should work on both side

### LINKED ISSUE #218: [FR] mobile Selection Scroll should work on both side
### Description

currently mobile selection only when selection start from top and user reach the current screen edge it start scroll downward
but it also scroll from bottom to top when top selection reach screen edge it should scroll automatically

### Impact

it will complete the mobile selection side

### Additional Context

_No response_

LucasXu0:
closes #218


LucasXu0:
FYI, @alihassan143

PR #221, opened 2023-06-20
feat: support IME in macOS

LucasXu0:


LucasXu0:
It seems that there is currently no way to simulate the IME.

PR #192, opened 2023-06-13
fix: copy handler migrated to latest api

alihassan143:
#189

PR #236, opened 2023-06-22
Revert "fix: copy handler migrated to latest api"

LucasXu0:
Reverts AppFlowy-IO/appflowy-editor#192


LucasXu0:
@alihassan143 This commit makes it impossible to copy words.

PR #244, opened 2023-06-23
feat: support customizing text attribute key and rendering

LucasXu0:
* support customizing the text attributes key

For example, we want to embed a page icon and page link into the paragraph.
1. customize a new attributes for delta, here's `{'mention': {'id': 'page_id'}}`
```dart
    final document = Document.blank()
      ..addParagraph(
        builder: (index) {
          return Delta()
            ..insert(
              '\$',
              attributes: {
                'mention': {'id': 'PAGE_ID'},
              },
            );
        },
      );
```
2. customize the textSpanDecorator function inside editor style, **VERY IMPORTANT: The length of the text must be 1 if you provide the WidgetSpan; otherwise, the selection will be inaccurate.**
```dart
final editorStyle = EditorStyle.desktop(
      // Example for customizing a new attribute key.
      textSpanDecorator: (textInsert, textSpan) {
        final attributes = textInsert.attributes;
        if (attributes == null) {
          return textSpan;
        }
        final mention = attributes['mention'] as Map?;
        if (mention != null) {
          return WidgetSpan(
            child: MouseRegion(
              cursor: SystemMouseCursors.click,
              child: GestureDetector(
                behavior: HitTestBehavior.translucent,
                child: Row(
                  mainAxisSize: MainAxisSize.min,
                  children: [
                    const Icon(Icons.edit_document),
                    Text(mention['id']),
                  ],
                ),
              ),
            ),
          );
        }
        return textSpan;
      },
    );
```
3. bingo.

Screenshot 2023-06-23 at 17 19 23


MichalNemec:
works flawlessly! When can we expect version with this included on pub.dev?

MichalNemec:
Coming back to it. It seems like whenever you do this and continue to type afterwards, its bugging out the text.

PR #259, opened 2023-06-28
fix: desktop tooltips incorrect

garv-shah:
While implementing the editor into my app, I noticed the desktop tooltips were incorrect:
SCR-20230628-ljdv
SCR-20230628-ljff
SCR-20230628-ljgs

As seen above, H2 and H3 still had the tooltip of H1, and the italics option had the tooltip of "bold". This is also reflected in the official AppFlowy app, but is a very quick and simple fix :D


LucasXu0:
LGTM.

PR #268, opened 2023-06-30
fix: slash menu keyword not shows up the option

johansutrisno:
linked to [#2844](https://github.com/AppFlowy-IO/AppFlowy/issues/2844)

[Bug] slash menu keyword


LucasXu0:
Hey, @johansutrisno. This PR causes a new issue. The menu won't dismiss even if there are no matched items already.

Screenshot 2023-07-03 at 11 02 04

I suggest you should dismiss the menu when there's no keyword match, maybe two or three times.

For example, 
1. HeadingX -> still show the menu
2. HeadingXX -> still show the menu
3. HeadingXXX -> not match again, dismiss the menu

johansutrisno:
okay @LucasXu0 , I will fix the issue

johansutrisno:
Hi @LucasXu0 , I have fixed the issue, this is the screenshot

https://github.com/AppFlowy-IO/appflowy-editor/assets/22239637/a10ab5d3-233b-4e96-80fe-e12a65ae61bc

LucasXu0:
@johansutrisno I found a new issue. If you type 'headingxx' and press the backspace key, the slash menu will dismiss.

lib/src/render/selection_menu/selection_menu_widget.dart
```diff
@@ -241,12 +246,16 @@ class _SelectionMenuWidgetState extends State<SelectionMenuWidget> {
...
-    if (keyword.length >= maxKeywordLength + 2) {
+    if (keyword.length >= maxKeywordLength + 2 &&
+        !(widget.deleteSlashByDefault && _searchCounter < 2)) {
       widget.onExit();
-    } else {
-      setState(() {
-        _showingItems = items;
-      });
+    }
+    setState(() {
```
Xazin, line 253: You removed the `else` here, wouldn't that allow `widget.onExit` to be called and then a `setState`. 

I would do `return widget.onExit();`

johansutrisno: ahh I see, thank you for remind me

johansutrisno:
Hi @LucasXu0 , I have fixed the the backspace key issue

https://github.com/AppFlowy-IO/appflowy-editor/assets/22239637/172a6d75-0115-48eb-9295-c37a94cd63ea

PR #272, opened 2023-07-01
feat: delete key functionality

MayurSMahajan:
Solves: [#335](https://github.com/AppFlowy-IO/AppFlowy/issues/335) and partially solves: [#2888](https://github.com/AppFlowy-IO/AppFlowy/issues/2888)

Added functionalities:

- [x] Delete the right character in the text.
- [x] Delete a selected node.
- [x] At the end of the line, pressing delete merges the next line into the current line.
- [ ] At the end of the line, pressing delete merges the next list node or checkbox node.


lib/src/editor/editor_component/service/shortcuts/command_shortcut_events/delete_command.dart
```diff
@@ -0,0 +1,110 @@
...
+
+/// Handle delete key event when selection is collapsed.
+CommandShortcutEventHandler _deleteInCollapsedSelection = (editorState) {
+  final selection = editorState.selection;
+  if (selection == null || !selection.isCollapsed) {
+    return KeyEventResult.ignored;
+  }
+
+  final position = selection.start;
+  final node = editorState.getNodeAtPath(position.path);
```
LucasXu0, line 36:
```dart
final delta = node?.delta;
if (node == null || delta == null) { return }

// use `delta` here  instead of `node.delta!`
```

lib/src/editor/editor_component/service/shortcuts/command_shortcut_events/delete_command.dart
```diff
@@ -0,0 +1,110 @@
...
+  final index = node.delta!.nextRunePosition(position.offset);
+  final transaction = editorState.transaction;
+  if (index == node.delta!.length) {
+    // merge the next node in the current node.
+    final nextNode = node.next;
+    if (nextNode == null) {
+      return KeyEventResult.ignored;
+    }
+    //TODO(Lucas): add logic for merging a bulletList or numberedList item.
+    if (nextNode.next == null && nextNode.children.isEmpty) {
```
LucasXu0, line 52: Why did you use `nextNode.next` here?

LucasXu0: Can you list all the cases so that I can implement it more effectively?

test/new/service/shortcuts/command_shortcut_events/delete_command_test.dart
```diff
@@ -0,0 +1,451 @@
...
+      // Welcome to AppFlowy Editor 🔥!|
+      //    Welcome to AppFlowy Editor 🔥!
+      // After
+      // Welcome to AppFlowy Editor 🔥!|Welcome to AppFlowy Editor 🔥!
+      //
+      //   test('''Delete the collapsed selection's next node when cursor is at end
+      //       and current node contains a delta
+      //       and the next node is the child of the current node''', () async {
+      //     final document = Document.blank().addParagraph(
+      //       initialText: text,
```
LucasXu0, line 128: Why comment out this code?

LucasXu0:
@MayurSMahajan Nice try! Can you list all the cases that related to `delete` key?

For example, 
1. collapsed selection
  1. at the begin, ...  
  2. at the middle, ...
  3. at the end, ... 
2. non collapsed selection, ...

test/new/service/shortcuts/command_shortcut_events/delete_command_test.dart
MayurSMahajan: This is the only functionality that needs to be implemented, I have commented this test, because the functionality is not yet implemented. We will just have to un-comment it and modify it a little bit in the future.

lib/src/editor/editor_component/service/shortcuts/command_shortcut_events/delete_command.dart
MayurSMahajan: I wrote this because in the backspace_command.dart we check the children and next node of a node when we want to append that node to the previous node. I guess it is a bit confusing. I tried to basically duplicate the behavior of the backspace_command.dart.

MayurSMahajan: In backspace handler, when we are at the start of a line and we press backspace, we want the current line to merge with the previous line.
Eg: 
Welcome to AppFlowy
|Welcome to AppFlowy
->
Welcome to AppFlowy | Welcome to AppFlowy

In delete handler, when we are at the end of a line and we press a delete, we want the next line to merge with the current line. 

Eg: 
Welcome to AppFlowy|
Welcome to AppFlowy
->
Welcome to AppFlowy | Welcome to AppFlowy

MayurSMahajan:
@LucasXu0 Here are the cases for the delete key:

Collapsed Selection:

1. at the begin and middle, delete the right character, implemented

2. at the end, (needs to be implemented)

- if the next node is TextNode: merge the next node in the current text node
- if the next node is a bullet list or checkbox, anything that has some text then, remove the bullet point/checkbox and just merge the text into the current text node.
- if the next node doesn't have any text, then ignore
- if the current node is not a text node then also ignore it.

Non-Collapsed Selection:

Just delete the selection entirely, implemented

LucasXu0:
> @LucasXu0 Here are the cases for the delete key:
> 
> Collapsed Selection:
> 
> 1. at the begin and middle, delete the right character, implemented
> 2. at the end, (needs to be implemented)
> 
> * if the next node is TextNode: merge the next node in the current text node
> * if the next node is a bullet list or checkbox, anything that has some text then, remove the bullet point/checkbox and just merge the text into the current text node.
> * if the next node doesn't have any text, then ignore
> * if the current node is not a text node then also ignore it.
> 
> Non-Collapsed Selection:
> 
> Just delete the selection entirely, implemented

Very clear. So do you need me to implement the missing part?

MayurSMahajan:
Yes, I can't wrap my head around it 😅, I guess if you can do it, that would be great.

Issue #71, opened 2023-04-13
[BUG] Add support for Query Schemes for Android and iOS (URL Launcher)


Xazin:
### Bug Description

URL Launcher requires some additional configuration for the example application, otherwise, it is not possible to open links in the editor.

Additionally, for anyone bringing in AppFlowy-Editor as a dependency, they should also configure URL Launcher for links to be openable. Thus we should consider adding this to the readme or to the documentation where appropriate.

See here for the solution: https://pub.dev/packages/url_launcher#configuration

### How to Reproduce

1. Run the `example` application on Android or iOS
2. Double-tap on a link
3. Link does not open

### Expected Behavior

1. Run the `example` application on Android or iOS
2. Double-tap on a link
3. Link opens

### Operating System

Android, iOS

### AppFlowy Editor Version(s)

main

### Additional Context

Logged when not configured:
```
I/UrlLauncher(28174): component name for http://appflowy.io is null
I/UrlLauncher(28174): component name for http://flutter.dev is null
```

PR #279, opened 2023-07-03
fix: add support for query schemes in example app

Xazin:
Additionally removes surfaceTintColor from AppBar, as it looks weird when body is behind appbar. Also resolves issue with _editorState not being initialized on reassemble in debugging.

Closes: #71


LucasXu0:
LGTM.

PR #291, opened 2023-07-05
fix: fix overflow in heading mobile toolbar menu

### LINKED ISSUE #285: A RenderFlex overflowed by 39 pixels on the right in Heading Toolbar Item.
### Bug Description

```
A RenderFlex overflowed by 39 pixels on the right.

The relevant error-causing widget was: 
  MobileToolbar MobileToolbar:.../my_editor.dart:58:11

The overflowing RenderFlex has an orientation of Axis.horizontal.
The edge of the RenderFlex that is overflowing has been marked in the rendering with a yellow and black striped pattern. This is usually caused by the contents being too big for the RenderFlex.
```

### How to Reproduce

In the mobile editor toolbar use `headingMobileToolbarItem`.

### Expected Behavior

It should either fit the mobile width or scroll horizontally.

### Operating System

IOS

1.0.4

### Screenshots

Simulator Screenshot - iPhone SE (3rd generation) - 2023-07-04 at 10 43 59

### Additional Context

_No response_

hyj1204:
To fix #285 
I refactored the UI to make sure it always has three buttons on the heading mobile toolbar menu.
If the width of the screen is too small, the button texts will show in 2 lines

Pasted Graphic

### DIFF
--- a/lib/src/editor/toolbar/mobile/toolbar_items/heading_mobile_toolbar_item.dart
+++ b/lib/src/editor/toolbar/mobile/toolbar_items/heading_mobile_toolbar_item.dart
@@ -47,7 +47,7 @@ class _HeadingMenuState extends State<_HeadingMenu> {
   @override
   Widget build(BuildContext context) {
     final style = MobileToolbarStyle.of(context);
-    final size = MediaQuery.of(context).size;
+    final size = MediaQuery.sizeOf(context);
     final btnList = headings.map((currentHeading) {
       // Check if current node is heading and its level
       final node =
--- a/lib/src/editor/toolbar/mobile/toolbar_items/heading_mobile_toolbar_item.dart
+++ b/lib/src/editor/toolbar/mobile/toolbar_items/heading_mobile_toolbar_item.dart
@@ -47,6 +47,7 @@ class _HeadingMenuState extends State<_HeadingMenu> {
   @override
   Widget build(BuildContext context) {
     final style = MobileToolbarStyle.of(context);
+    final size = MediaQuery.of(context).size;
     final btnList = headings.map((currentHeading) {
       // Check if current node is heading and its level
       final node =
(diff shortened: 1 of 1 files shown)


alihassan143:
> To fix #285 I refactored the UI to make sure it always has three buttons on the heading mobile toolbar menu. If the width of the screen is too small, the button texts will show in 2 lines
> 
> Pasted Graphic

instead of showing in 2 lines use wrap widget if screen is too small third heading will be rendered at second line that is more better looking than rendering text in 2 lines

hyj1204:
@alihassan143 

The current layout rules in toolbar menu on mobile are: 

- The buttons on the toolbar menu take up all the space of the width of the screen
- For every row will lay 2 or 3 buttons based on which toolbar item it is
- Every button has **the same space** between them, and the rest of the space will be used to render buttons. 

The main purpose of it is to make all components(buttons) take all the width without leaving some undefined width of spacing. In this way, the width of the screen control how wide it is for a button, it will look like this on most screen sizes. 
image
If the width of screen is not wide enough to let the text inside button to show within one row, it will look like this
image


The way you suggest: 

If we change to wrap, we have to rely on the text inside button to decide how wide the button it is, we will have no control over how much space will be left in a row.  
On a wide screen:
Screenshot 2023-07-06 at 10 05 33 AM
On a narrow screen: 
image

The way I implemented will fit most sizes of screens without unknown width spacing, which follows the same layout rules in other toolbar items menus on mobile.

Xazin:
I am wondering if it would be fine to hide text where the icons themselves are so common that the user wouldn't have a problem interpreting it, on small screens.

lib/src/editor/toolbar/mobile/toolbar_items/heading_mobile_toolbar_item.dart
```diff
@@ -47,45 +47,54 @@ class _HeadingMenuState extends State<_HeadingMenu> {
   @override
   Widget build(BuildContext context) {
     final style = MobileToolbarStyle.of(context);
+    final size = MediaQuery.of(context).size;
```
Xazin, line 50:
```suggestion
    final size = MediaQuery.size;
```
I believe this should also be possible.

Xazin (approved):

lib/src/editor/toolbar/mobile/toolbar_items/heading_mobile_toolbar_item.dart
LucasXu0: `MediaQuery.size`? How to get the size without contenxt?

Xazin: It's my bad. I wanted to write `sizeOf(context)`.

LucasXu0:
@hyj1204 Please check the comments.

hyj1204:
> I am wondering if it would be fine to hide text where the icons themselves are so common that the user wouldn't have a problem interpreting it, on small screens.

Yes, for the heading part, the icons are easy to understand, I just feel we should try to keep the same UI style rules in all the mobile toolbar item menus. If we change all the buttons in the toolbar item menu to show icons only, I am not sure if all icons are easy to understand, also the icons will be centered in that case I think.

PR #297, opened 2023-07-08
feat: refactor image block component

rileyhawk1417:
- Refactoring the widget tree into smaller classes.
- This PR is meant to replace the old  [PR](https://github.com/AppFlowy-IO/appflowy-editor/pull/195)


Chandu-bs:
I think this could resolve the cursor blinking beside the image and getting editor focus after the image upload

```
extension InsertImage on EditorState {
  Future<void> insertImageNode(
    String src,
  ) async {
    final selection = this.selection;
    if (selection == null || !selection.isCollapsed) {
      return;
    }
    final node = getNodeAtPath(selection.end.path);
    if (node == null) {
      return;
    }
    final transaction = this.transaction;
    // if the current node is empty paragraph, replace it with image node
    if (node.type == ParagraphBlockKeys.type &&
        (node.delta?.isEmpty ?? false)) {
      if (selection == null || !selection.isCollapsed) {
        return;
      }
      final path = selection.end.path;
      final delta = node?.delta;
      if (node == null || delta == null) {
        return;
      }
      final insertedPath = delta.isEmpty ? path : path.next;
      final transactions = transaction
        ..insertNode(
          insertedPath,
          imageNode(
            url: src,
          ),
        )
        ..insertNode(insertedPath, paragraphNode())
        ..afterSelection = Selection.collapse(insertedPath.next, 0);
      return apply(transactions);
    } else {
      transaction.insertNode(
        node.path.next,
        imageNode(
          url: src,
        ),
      );
      return apply(transaction);
    }
  }
}
```

rileyhawk1417:
@Chandu-bs will test it out

PR #300, opened 2023-07-11
chore: support customize the font family for bold

LucasXu0:


jessicamrbr:
Wow I think this might be useful to combine with other tests I was trying

https://github.com/google/material-design-icons/issues/1543

Issue #311, opened 2023-07-13
[Bug] the background color of the block component is overflow


LucasXu0:
### Bug Description

I found the background of the child block component is overflow when integrating the latest version into AppFlowy.

<img width="1015" alt="Screenshot_2023-07-13_at_14 28 27" src="https://github.com/AppFlowy-IO/appflowy-editor/assets/11863087/07f239aa-7c9e-4815-bf00-aa752fc40cfc">

### How to Reproduce

1. change the background color of the to-do list block.

### Expected Behavior

<img width="993" alt="Screenshot 2023-07-13 at 18 39 07" src="https://github.com/AppFlowy-IO/appflowy-editor/assets/11863087/ead3fbbf-35e6-462b-a1c7-9ba55fb9dc8c">

### Operating System

All Platform

### AppFlowy Editor Version(s)

1.2.0

Issue #263, opened 2023-06-29
[FR] AppFlowy Plugins in AppFlowy Editor


garv-shah:
### Description

The AppFlowy Editor plugin system is great, but there are a lot of plugins in the AppFlowy app that aren't included by default in the AppFlowy Editor package. This would be particularly useful, as many design decisions in AppFlowy Editor are specifically made to accomodate plugins from the main app.
For example, on the line https://github.com/AppFlowy-IO/appflowy-editor/blob/ed75bf61df44281f2605433f8a59460f7417bbc1/lib/src/editor/editor_component/service/renderer/block_component_action.dart#L4 there is a spacing of 50 pixels, which accommodates the action menu plugin in the AppFlowy app, but creates issues when developing plugins solely with AppFlowy Editor, as it creates a 50 pixel gap seemingly out of nowhere. The easiest fix for this seems to just be to use the same plugin that the AppFlowy app does, hence the suggestion of these being built into the editor package.
 
Currently, the seemingly fastest way to get the functionality from the real app into the editor is to copy many of the files from AppFlowy into your own project. It would be quite nice if it was possible to enable some more of these plugins from the default editor, making it easier to implement functionality on par with the main app by using the editor.

### Impact

This change mostly helps developers who would like to implement the features already included in the main AppFlowy app, but are not present in the editor.

LucasXu0:
Let me see how to refactor it and provide a more customizable way.

garv-shah:
thank you!

PR #312, opened 2023-07-13
fix: block selection overflow

LucasXu0:
closes https://github.com/AppFlowy-IO/appflowy-editor/issues/311


LucasXu0:
https://github.com/AppFlowy-IO/appflowy-editor/issues/263#issue-1780444561

And removed the limited number mentioned in the above link.

Issue #327, opened 2023-07-19
[FR] Markdown command shortcut for toggle highlighting


MayurSMahajan:
### Description

`Ctrl/Cmd + Shift + H` should toggle the highlight attribute on text.

### Impact

Everyone

PR #333, opened 2023-07-20
feat: toggle highlight using ctrl/cmd + shift + h

MayurSMahajan:
Solves: #327


LucasXu0:
The logic does not seem correct. We should cancel the highlight color if it is not equal to the `defaultHighlightColor`.

MayurSMahajan:
@Xazin yes you can help me, review the PR and let me know of any improvements

lib/src/editor/editor_component/service/shortcuts/command_shortcut_events/toggle_colors_command.dart
```diff
@@ -0,0 +1,70 @@
...
+  EditorState editorState, {
+  required ToggleColorsStyle style,
+}) {
+  if (PlatformExtension.isMobile) {
+    assert(false, 'toggle highlight is not supported on mobile platform.');
+    return KeyEventResult.ignored;
+  }
+
+  final selection = editorState.selection;
+  if (selection == null) {
```
Xazin, line 49:
```suggestion
  if (selection == null || selection!.isCollapsed) {
```
We don't need to do anything if it's collapsed right?

lib/src/editor/editor_component/service/shortcuts/command_shortcut_events/toggle_colors_command.dart
```diff
@@ -0,0 +1,70 @@
...
+    return KeyEventResult.ignored;
+  }
+
+  //check if already highlighted
+  final nodes = editorState.getNodesInSelection(selection);
+  final isHighlight = nodes.allSatisfyInSelection(selection, (delta) {
```
Xazin, line 55:
```suggestion
  final isHighlighted = nodes.allSatisfyInSelection(selection, (delta) {
```

lib/src/editor/editor_component/service/shortcuts/command_shortcut_events/toggle_colors_command.dart
```diff
@@ -0,0 +1,70 @@
...
+    return delta.everyAttributes(
+      (attributes) => attributes[AppFlowyRichTextKeys.highlightColor] != null,
```
Xazin, line 57: I believe what Lucas wanted you to do, is:

- If the highlight is already `== style.highlightColor` then do nothing
- If the highlight is another color then set to `style.highlightColor`
- If there is no highlight then set to `style.highlightColor`.

So this check needs to be changed.

lib/src/editor/block_component/standard_block_components.dart
```diff
@@ -111,6 +111,7 @@ final List<CommandShortcutEvent> standardCommandShortcutEvents = [
   //
   toggleTodoListCommand,
   ...toggleMarkdownCommands,
+  ...toggleColorCommands(),
```
Xazin, line 114: Here we need to provide the highlight color which is the `default highlight color`. Part of Lucas' comment.

lib/src/editor/editor_component/service/shortcuts/command_shortcut_events/toggle_colors_command.dart
MayurSMahajan: cc @LucasXu0

Xazin (approved):

PR #341, opened 2023-07-25
Fix Bug 339

### LINKED ISSUE #339: [Bug] Desktop color picker issue
### Bug Description

1、desktop color picker Display screen at specific location
2、Custom color Difficulty in triggering editing
3、Custom color Hex value display error

### How to Reproduce

Simply select the text color operation

### Expected Behavior

The operation floating window is fully displayed on the screen and the operation display is normal

### Operating System

desktop

master

### Screenshots

image
image

### Additional Context

_No response_

q200892907:
closes #339


lib/src/editor/util/color_util.dart
```diff
@@ -39,6 +39,6 @@ extension ColorExtension on String {
 
 extension HexExtension on Color {
   String toHex() {
-    return value.toRadixString(16);
+    return '0x${value.toRadixString(16)}';
```
LucasXu0, line 42: why adding `0x` here?

q200892907: Because the parameter structure required for ColorOption's colorHex is in the 0xFFFFFF format, and toHex does not add 0x, resulting in a 2-format error when ColorPicker selects and customizes colors

lib/src/editor/toolbar/desktop/items/color/color_menu.dart
```diff
@@ -23,7 +23,7 @@ void showColorMenu(
   final offset = rect.center;
   final editorOffset = editorState.renderBox!.localToGlobal(Offset.zero);
   final editorHeight = editorState.renderBox!.size.height;
-  final threshold = editorOffset.dy + editorHeight - 200;
+  final threshold = editorOffset.dy + editorHeight - 250;//use color_picker_height
```
Xazin, line 26: Remove comment or add comment above threshold.. (formatting)

lib/src/editor/util/color_util.dart
Xazin: Depending on our usage, a `Color` can be constructed using the 32 bit value, `Color(value)`.

So if we don't display the hex code anywhere, and it's just for storing and reading a Color, I'd prefer using `Color.value` over this hacky method.

q200892907: Selecting a color first and then going to the custom color view may cause numerical display issues

lib/src/editor/toolbar/desktop/items/color/color_menu.dart
q200892907: OK

PR #354, opened 2023-08-03
feat: mobile gesture detector improvement

### LINKED ISSUE #308: [FR] Support drag selection handlers - When long pressing a word it is selected, it should be possible to drag the left or right part of the selection to expand it

hyj1204:
- [x] Separate mobile and desktop selection gesture detector
- [x] Refactor MobileSelectionGestureDetector to achieve the following interaction
    - [x] 1. onTap: Show the cursor in the document
    - [x] 2. onDoubleTap: Select the word and show draggable handlers to expand the selected area. 
    ~~3. onTripleTap: Select the node(block) and show draggable handlers to expand the selected area.~~
    **Move to another PR, along with implementing customized GestureDetector(by RawGestureDetector)**
- [x] ~~Remove `onPanUpdate` on `MobileSelectionServiceWidget`, which means when the user moves the pan, it won't do anything like select texts. I removed it because users usually scroll a lot on the phone, it is hard to tell if users want to select texts or just scroll on the screen.~~ 
To clarify the above texts, I removed the code related to selecting text by mouse(It was in `onPanUpdate`) in mobile since we don't need it. The code in `onPanUpdate` right now is for changing the selection area by user's dragging behavior. 
In mobile, the only way to select certain texts will be:
First, double tap to select the word, then handlers will show around it. 
Then, move the handlers to expand the selection area.
This feature still has a bug right now. After the first time moving the left handler, it stop working if we want to move it again. This problem only exists in the left handler,  the right handler works totally fine. I plan to fix it in another PR. 

https://github.com/AppFlowy-IO/appflowy-editor/assets/14248245/eb0c16ca-6957-4285-aed3-743b1741284e
- [X] Fix #308 
**I implemented double tap instead of long press to select a word for now.**

~~- [ ] When some texts are selected, the selection area is unclickable. Users only can update the selection area by tapping on the rest area or dragging the handler to expand it.~~
I tried `IgnorePointer` and `AbsorbPointer` but it doesn't work. The desktop version is still clickable for the selection area, so I removed it from this PR for now.

- [X] Add moving cursor feature

https://github.com/AppFlowy-IO/appflowy-editor/assets/14248245/b87c0a63-ae23-41a7-80a6-0f52badca0dd

 ---
**NOTE:**
I have been working on this PR for a while. I'd like to merge the finished features first and then add more features or improvements later. 

In this PR, I refactored `MobileSelectionGestureDetector` by build-in `GestureDetector` to see if all the Gesture**Callback is working well. In order to understand the code better, here is the trigger order for every gesture that `GestureDetector` provides. I used this order to 1) record where user click/point at 2) change the selection in editorState accordingly, and save some `Rect` to render the selection effect later
Tap: 1)  onTapUp =>  2) onTap
DoubleTap: 1) onDoubleTapDown => 2) onDoubleTap
Dragging: 1) onPanDown => 2) onPanUpdate


hyj1204:
This PR is closed because its features have been implemented by  #481. 
The key commit is [here](https://github.com/AppFlowy-IO/appflowy-editor/pull/481/commits/538db46b1a2d8ecf2e29a7f2ca825efacdb0732d)  which uses `onTapUp` instead of `onTapDown` in the previous code. The code inside `onTapDown`  didn’t get called because `onPanStart` won over/blocked it, so it didn’t work.

It still has a related bug here #491

PR #358, opened 2023-08-05
fix: remove flutter overlay

Xazin:
Tested and couldn't find any issues


LucasXu0:
https://github.com/AppFlowy-IO/appflowy-editor/assets/11863087/32a4775f-b9f1-4783-8ca8-6a160d88b5ef

@Xazin found a new issue.

Xazin:
> x.mp4 
> @Xazin found a new issue.

This happens to me even with the old version of the `overlay.dart`.

PR #365, opened 2023-08-10
fix: lose selection after pasting

LucasXu0:


lib/src/service/internal_key_event_handlers/copy_paste_handler.dart
```diff
@@ -135,6 +149,38 @@ void handleCopy(EditorState editorState) async {
...
+    final length = insertedNode.delta?.length ?? 0;
+    transaction.afterSelection = Selection.collapse(selection.end.path, length);
+    editorState.apply(transaction);
+  } else {
+    transaction.insertTextDelta(
+      node,
+      offset,
+      insertedDelta,
+    );
+    editorState.apply(transaction);
```
Xazin, line 180: No `afterSelection` here?

LucasXu0: Basically, we manage the `afterSelection` of the text-related operations inside the functions. e.g. `insertText`, `deleteText`.

LucasXu0: Certainly, we can discuss and optimize them later. My plan is to bump the version now and improve them in the upcoming release.

Issue #370, opened 2023-08-12
[Bug] Divider Throwing Error


Mukund-Tandon:
### Bug Description

![ezgif com-video-to-gif (3)](https://github.com/AppFlowy-IO/appflowy-editor/assets/71614009/73337da7-67cb-4f78-82b5-ca9133a6f69d)
In the above GIF we can see the when we type three ‘-’ it create a divider but if we press a fourth ‘-’ it will through an error

### How to Reproduce

On any line type '-' 3 times on forth time it throws error .

### Expected Behavior

After making divider cursor should move to next line like if we insert divider from slash menu then the cursor comes to next line 

### Operating System

MacOs

### AppFlowy Editor Version(s)

1.2.3

Mukund-Tandon:
I have fixed the issue , will just write the tests for it

PR #371, opened 2023-08-12
fix: fixed the cursor not moving to next line after inserting divider

Mukund-Tandon:
fixes: #370 
Before:
ezgif com-video-to-gif (3)

After -

https://github.com/AppFlowy-IO/appflowy-editor/assets/71614009/0f0a2de1-533a-4356-8337-48323935b081

### DIFF
--- a/lib/src/editor/block_component/divider_block_component/divider_block_component.dart
+++ b/lib/src/editor/block_component/divider_block_component/divider_block_component.dart
@@ -96,13 +96,13 @@ class _DividerBlockComponentWidgetState
       child: child,
     );
 
-    if (widget.showActions && widget.actionBuilder != null) {
-      child = BlockComponentActionWrapper(
-        node: node,
-        actionBuilder: widget.actionBuilder!,
-        child: child,
-      );
-    }
+    // if (widget.showActions && widget.actionBuilder != null) {
+    //   child = BlockComponentActionWrapper(
+    //     node: node,
+    //     actionBuilder: widget.actionBuilder!,
+    //     child: child,
+    //   );
+    // }
 
     return child;
   }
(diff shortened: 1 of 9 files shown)


MayurSMahajan (changes requested):
Great work! Can you add a test to assure this functionality, so that if it breaks in the future, we would know.

Mukund-Tandon:
> Great work! Can you add a test to assure this functionality, so that if it breaks in the future, we would know.

Ok,I will add them

Xazin (approved):

lib/src/editor/block_component/divider_block_component/divider_block_component.dart
```diff
@@ -96,13 +96,13 @@ class _DividerBlockComponentWidgetState
...
-        child: child,
-      );
-    }
+    // if (widget.showActions && widget.actionBuilder != null) {
+    //   child = BlockComponentActionWrapper(
+    //     node: node,
+    //     actionBuilder: widget.actionBuilder!,
+    //     child: child,
+    //   );
+    // }
```
LucasXu0, line 105: Why do you comment out this code?

lib/src/editor/block_component/divider_block_component/divider_block_component.dart
```diff
@@ -120,7 +120,7 @@ class _DividerBlockComponentWidgetState
   bool get shouldCursorBlink => false;
 
   @override
-  CursorStyle get cursorStyle => CursorStyle.cover;
+  CursorStyle get cursorStyle => CursorStyle.borderLine;
```
LucasXu0, line 123: Don't modify it

lib/src/editor/block_component/divider_block_component/divider_block_component.dart
Mukund-Tandon: I am sorry ,forgot to uncomment it , I was just testing and understanding the code base

MayurSMahajan (approved):

Issue #380, opened 2023-08-14
[Bug]: Must be some hint/label text for using editor


yatendra2001:
### Bug Description

We can write some hint text for the user to prompt `'/' for commands`

![Screenshot 2023-08-14 at 12.38.08 PM\|690x284](https://global.discourse-cdn.com/standard11/uploads/appflowy/optimized/1X/aaa448f3357bf24953877d60755651fbf6b1f630_2_1380x568.png)

### How to Reproduce

--

### Expected Behavior

<img width="682" alt="Screenshot 2023-08-14 at 12 43 30 PM" src="https://github.com/AppFlowy-IO/appflowy-editor/assets/62821607/83dc43eb-dd36-4697-b587-97fa7c82fa3f">

### Operating System

MacOS

### AppFlowy Editor Version(s)

main

PR #390, opened 2023-08-16
fix: placeholder on paragraph when selected

Xazin:
Closes: #380

# Preview

https://github.com/AppFlowy-IO/appflowy-editor/assets/42929161/f4b68051-28d6-4f76-ac66-cc6004de3577

### DIFF
--- a/lib/src/editor/block_component/text_block_component/text_block_component.dart
+++ b/lib/src/editor/block_component/text_block_component/text_block_component.dart
@@ -106,33 +106,21 @@ class _TextBlockComponentWidgetState extends State<TextBlockComponentWidget>
   @override
   void initState() {
     super.initState();
-    editorState.selectionService.currentSelection
-        .addListener(_onSelectionChange);
+    editorState.selectionNotifier.addListener(_onSelectionChange);
     _onSelectionChange();
   }
 
   @override
   void dispose() {
-    editorState.selectionService.currentSelection
-        .removeListener(_onSelectionChange);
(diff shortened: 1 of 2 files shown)


lib/src/editor/block_component/text_block_component/text_block_component.dart
```diff
@@ -101,6 +101,41 @@ class _TextBlockComponentWidgetState extends State<TextBlockComponentWidget>
...
+
+      final selectedNodes = editorState.getNodesInSelection(selection);
+      if (selectedNodes.length == 1 && selectedNodes.contains(node)) {
+        _showPlaceholder = true;
+        return;
+      }
+
+      _showPlaceholder = false;
+    });
+  }
```
LucasXu0, line 137:
```suggestion
  @override
  void initState() {
    super.initState();
    editorState.selectionNotifier.addListener(_onSelectionChange);
    _onSelectionChange();
  }

  @override
  void dispose() {
    editorState.selectionNotifier.removeListener(_onSelectionChange);
    super.dispose();
  }

  void _onSelectionChange() {
    setState(() {
      final selection = editorState.selection;
      _showPlaceholder = selection != null &&
          (selection.isSingle && selection.start.path.equals(node.path));
    });
  }
```

LucasXu0: The `selectionService` should not be used in this case as it will raise an error when not in a UI context.

PR #392, opened 2023-08-17
Title Node Plugin For AppFlowy Editor

rileyhawk1417:
This plugin has been migrated from the [AppFlowy repo](https://github.com/AppFlowy-IO/AppFlowy/pull/2078)

The original issue can be found [here](https://github.com/AppFlowy-IO/AppFlowy/issues/1774)


Xazin:
Hey @LucasXu0 can you help take a look at this?

I'm not a fan of the implementation itself, maybe you have a better method in mind?

Xazin:
@LucasXu0 

I see two issues still here, what is your take on them:
1. I'm not a fan of the way we're currently enforcing the title block to always be the first node in a document, do you have some ideas for another route I could look at for improving this?
2. Currently it's possible to select and add attributes to a title, I assume we should disallow this and only allow for plain text / UTF emojis. Since formatting as a list or quote, breaks the plugin.

Xazin:
> @LucasXu0
> 
> I see two issues still here, what is your take on them:
> 
> 1. I'm not a fan of the way we're currently enforcing the title block to always be the first node in a document, do you have some ideas for another route I could look at for improving this?
> 2. Currently it's possible to select and add attributes to a title, I assume we should disallow this and only allow for plain text / UTF emojis. Since formatting as a list or quote, breaks the plugin.
> 
> Once we can merge this, I can continue work on AppFlowy to remove the title bar and abstract the share button and font settings out of the title bar as well.

Will address the 2nd issue and then merge, to continue with the improvements on AppFlowy.

PR #395, opened 2023-08-18
feat: double hyphen to em dash

Xazin:


Xazin:
Closes https://github.com/AppFlowy-IO/AppFlowy/issues/3076

PR #404, opened 2023-08-22
fix: exception on prev node text direction null

zoli:
The code was throwing exception on changing the current node text
direction to auto when the previous node had a text direction attribute
with null value.
Also changed to only use direct previous node text direction and
direct parent for inheriting text direction when the current node
text direction is auto and undeterminable (e.g. empty text).

This should get merged after #402.


LucasXu0:
@zoli I have merged https://github.com/AppFlowy-IO/appflowy-editor/pull/402 already. Please resolve the conflicts.

zoli:
> @zoli I have merged #402 already. Please resolve the conflicts.

Done.

Issue #396, opened 2023-08-18
[Bug] Find and Replace fails on certain usecase


squidrye:
### Bug Description

Find and replace fails when words to be replaced are placed one after another, sometimes this kind of replacement would freeze the entire editor, user won’t be able to work unless they restart the editor, I couldn’t reproduce the scenario where editor froze with every test but it happens quite frequently.

### How to Reproduce

1. Type in following sequence of word
> aaaa
2. Find ‘a’
3. Replace with word of length > 1
4. For eg if replaced by ‘test’
> tttteststststaaa

### Expected Behavior

found words should be correctly replaced by target word

### Operating System

Arch Linux

### AppFlowy Editor Version(s)

1.2.2

### Screenshots

![ezgif com-video-to-gif (1)](https://github.com/AppFlowy-IO/appflowy-editor/assets/84044317/66d3fde6-c19e-4582-8eb4-8ec218bbe445)

Xazin:
Hey @MayurSMahajan will you work on this?

The bug occurs from inside `replaceSelectedWord`.

MayurSMahajan:
I will work on this. I was on a vacation for a couple of days.

PR #418, opened 2023-08-26
fix: replace matches on the same node

MayurSMahajan:
Solves: #396 

https://github.com/AppFlowy-IO/appflowy-editor/assets/47064215/a31683ec-a10d-4a6a-b84b-6188a54c7e7e


LucasXu0:
LGTM

LucasXu0 (approved):

PR #419, opened 2023-08-28
feat: bulk open links shortcut

Tony-MK:
Created a keyboard shortcut to bulk open selected links in a document. 
Reference issue: [#2323](https://github.com/AppFlowy-IO/AppFlowy/issues/2323)

### DIFF
--- a/lib/src/editor/editor_component/service/shortcuts/command_shortcut_events/open_links_command.dart
+++ b/lib/src/editor/editor_component/service/shortcuts/command_shortcut_events/open_links_command.dart
@@ -1,0 +1,45 @@
+import 'package:appflowy_editor/appflowy_editor.dart';
+import 'package:flutter/material.dart';
+
+/// Option/Alt + Shift + Enter: to open links
+/// - support
+///   - desktop
+///   - web
+///
+final CommandShortcutEvent openLinksCommand = CommandShortcutEvent(
+  key: 'open links',
+  command: 'alt+shift+enter',
+  handler: _openLinksHandler,
+);
+
+KeyEventResult _openLinksHandler(
+  EditorState editorState,
+) {
+  if (PlatformExtension.isMobile) {
+    assert(false, 'open links is not supported on mobile platform.');
+    return KeyEventResult.ignored;
+  }
+
+  final selection = editorState.selection;
+  if (selection == null || selection.isCollapsed) {
+    return KeyEventResult.ignored;
+  }
+
+  final nodes = editorState.getNodesInSelection(selection);
+
+  // A set to store the links which have been opened
+  // to prevent opening new tabs for the same link
+  Set<String> openedLinks = {};
+
+  for (final node in nodes) {
+    for (final operation in node.delta!) {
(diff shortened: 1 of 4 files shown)


LucasXu0:
@Tony-MK please check the tests. some of them failed.

Tony-MK:
Thanks @LucasXu0, I am currently trying to fix them, then it will be ready for review.

lib/src/editor/editor_component/service/shortcuts/command_shortcut_events/open_links_command.dart
```diff
@@ -0,0 +1,45 @@
...
+
+  for (final node in nodes) {
+    for (final operation in node.delta!) {
+      final link = operation.attributes?[BuiltInAttributeKey.href];
+      if (link == null || openedLinks.contains(link)) continue;
+
+      openedLinks.add(link);
+      safeLaunchUrl(link);
+    }
+  }
```
LucasXu0, line 42:
```suggestion
  // A set to store the links which have been opened
  // to prevent opening new tabs for the same link
  final openedLinks = <String>{};

  for (final node in nodes) {
    final delta = node.delta;
    if (delta == null) {
      continue;
    }

    // Get all links in the node
    final links = delta
        .map<String?>((op) => op.attributes?[AppFlowyRichTextKeys.href])
        .whereNotNull()
        .toSet()
        .difference(openedLinks);

    for (final link in links) {
      safeLaunchUrl(link);
    }

    openedLinks.addAll(links);
  }
```

LucasXu0: - don't force unwrap the delta value.
- optional: use the chained function to get the links.

LucasXu0: @Tony-MK Please commit the suggestions or do you have any other better ideas?

Tony-MK:
Sorry for the wait, @LucasXu0. I don't think I have any better ideas. Your suggestions are good. 
However, I do have a question regrading the best time to call the code snippet below. 

`
for (final link in links) {
    safeLaunchUrl(link);
}
`

What difference does it make if the function loops through `openedLinks` and called after the initial loop has finished? Despite the `openedLinks` set having more items than a node's links set, the snippet will only be called once so I believe it should be faster and also probably be prettier to read. Can you please explain for me why, if I am wrong?

LucasXu0:
@Tony-MK You're right. It would be better to call the "open link" function after the loop.

```
  // A set to store the links which should be opened
  final links = nodes
      .map((e) => e.delta)
      .whereNotNull()
      .expand(
        (e) => e.map<String?>(
          (op) => op.attributes?[AppFlowyRichTextKeys.href],
        ),
      )
      .whereNotNull()
      .toSet();

  for (final link in links) {
    safeLaunchUrl(link);
  }
```

Tony-MK:
Thanks @LucasXu0, I will commit the changes ASAP.

Issue #386, opened 2023-08-14
[Bug] Confusing UX on Replace Dialog after pressing Enter key


MayurSMahajan:
### Bug Description

While replacing a word with Replace Dialog in the AppFlowy Editor,
we are able to replace the first match by pressing Enter key
But then the focus shifts into the Editor and now pressing the Enter key inserts a newline character instead of replacing the next match.

### How to Reproduce

1. Open AppFlowy Editor
2. Open Replace Dialog
3. Enter a find string and a replace string.
4. Now focusing on the Replace string press Enter
5. Now press enter again.

### Expected Behavior

Two behaviors can be expected:

1. The next matched word should be replaced.
2. It should not be possible to replace any match with Enter key in first place

### Operating System

Windows 11

### AppFlowy Editor Version(s)

1.2.3

### Screenshots

![replace_ux](https://github.com/AppFlowy-IO/appflowy-editor/assets/47064215/99c2934c-05ef-4e87-afd3-08ea26bb46a9)

MayurSMahajan:
@LucasXu0 @Xazin 
Guys, which behavior should we expect? I am leaning towards 1.
Also assign this issue to me.

Xazin:
The problem here is the editor receives focus and thus the enter is not registered from the find dialog.

Resolve that and it will be fine I believe.

PR #440, opened 2023-09-06
fix: request focus in find replace menu

Xazin:
This PR includes a workaround to keep focus after submitting in the find/replace text fields, otherwise the focus is forced on the editor due to the interactions required.


Xazin:
Closes #386

PR #471, opened 2023-09-16
fix: resolve "Bad state: no element" when in-table paragraph deleted

### LINKED ISSUE #465: [Bug] deleting tables using DEL button causes render error.
### Bug Description

When deleting the table I get an error table no element. Instead of backspace I pressed DEL button.
image

### How to Reproduce

1. git clone project
2. run on windows
image
3. press enter two times
image
4. add table 
image
5. press on empty space before the table
image
6. press DEL button (not the backspace button)
image

### Expected Behavior

Expected behavior is to delete the table in my opinion.

### Operating System

windows 10

1.3.0

### Screenshots

_No response_

### Additional Context

_No response_

sun-jiao:
Close #465 .

This occurred because the table was not deleted, instead, the paragraph in the first cell was deleted. As a result, no child elements could be found in the cell when building the table (and thus throws "bad state, no element").

I'm not sure if this is the right way. If this fix is not correct or not the best practice, please make any suggestions or modifications, or, of course, close it.

### DIFF
--- a/lib/src/extensions/node_extensions.dart
+++ b/lib/src/extensions/node_extensions.dart
@@ -132,6 +132,14 @@ extension NodeExtensions on Node {
     }
     return false;
   }
+
+  Node? findParent(bool Function(Node element) test) {
+    if (test(this)) {
+      return this;
+    }
+    final parent = this.parent;
+    return parent?.findParent(test);
+  }
 }
 
 extension NodesExtensions<T extends Node> on List<T> {
--- a/lib/src/extensions/node_extensions.dart
+++ b/lib/src/extensions/node_extensions.dart
@@ -79,33 +79,25 @@ extension NodeExtensions on Node {
   }
 
   // find the node from it's children or it's next sibling to find the node that matches the given predicate
-  Node? findDownward(bool? Function(Node element) test) {
+  Node? findDownward(bool Function(Node element) test) {
     final children = this.children.toList();
     for (final child in children) {
-      switch (test(child)) {
-        case null:
-          return null;
-        case true:
-          return child;
-        case false:
-          if (child.children.isNotEmpty) {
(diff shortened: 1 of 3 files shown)


LucasXu0:
@sun-jiao the CI analyzer failed.

sun-jiao:
@LucasXu0  fixed

lib/src/extensions/node_extensions.dart
```diff
@@ -79,25 +79,33 @@ extension NodeExtensions on Node {
   }
 
   // find the node from it's children or it's next sibling to find the node that matches the given predicate
-  Node? findDownward(bool Function(Node element) test) {
+  Node? findDownward(bool? Function(Node element) test) {
```
LucasXu0, line 82: Defining the return type as an optional bool type seems unclear.

I believe what you need is:

1. If the return value is null, it means to skip the rest of the test.
2. If it returns false, continue searching deeply.
3. If it returns true, you've found what you're looking for.

Correct me if it's wrong.

LucasXu0: If so, I suggest you define the return like that.

enum {
 accept, // yes
 reject,  // no
 skipRemaining, // null 
}

sun-jiao: @LucasXu0 

Yes that's right. While my concern is whether it's the best fix for the problem to add a skip option here. And whether the same edits should be made to `lastChildWhere` and `previousNodeWhere`, although they don't seem to be needed at the moment.

sun-jiao:
@LucasXu0 I find another way to fix it:

Check if the current node and next/prev node are both outside of table or in the same table. 

If only one of them is in a table or they're in different tables, the deletion will be terminated.

PR #488, opened 2023-09-21
fix: required named parameter 'onLiveTextInput' must be provided

LucasXu0:


sun-jiao:
I think we'd better add more flutter versions to ./github/workflows/test.yaml to avoid further similar errors.

PR #489, opened 2023-09-21
fix: hide placeholder in table to avoid strange behaviour

sun-jiao:
The placeholder in table cells cause very strange behaviour.
And the slash command, in fact, does not work in table cells, so the placeholder is totally redundant. 
  
[录屏 2023-09-21 10-22-39.webm](https://github.com/AppFlowy-IO/appflowy-editor/assets/14086980/bebb328a-2758-4be1-8435-6bbb96c18233)


LucasXu0:
Please don't include the code related to the table within a paragraph block.

A workaround.

```
    map[ParagraphBlockKeys.type] = TextBlockComponentBuilder(
      configuration: BlockComponentConfiguration(
        placeholderText: (node) {
          // disable the placeholder for table block
          Node? parent = node.parent;
          while (parent != null) {
            if (parent.type == TableBlockKeys.type) {
              return '';
            }
            parent = parent.parent;
          }
          return AppFlowyEditorLocalizations.current.slashPlaceHolder;
        },
      ),
    );
```

sun-jiao:
@LucasXu0 Do you mean configure it in my own project rather than in the library?

LucasXu0:
> @LucasXu0 Do you mean it should be configured in my own project rather than in the library?

Yes, for example, the placeholder currently uses 'enter a / ...,' and you disable it within the text block based on a specific condition. Then, if one day the placeholder changes to 'type something...' and the developers need to display the placeholder within the table block, it won't function as intended.

PR #490, opened 2023-09-21
fix: fix mobile selection drag

dggbond:
bug screen record:
when you drag left handler first, then you drag right handler will cause this bug, selection changed wrong.

bug

fixed:
fixed

### DIFF
--- a/lib/src/editor/editor_component/service/selection/mobile_selection_service.dart
+++ b/lib/src/editor/editor_component/service/selection/mobile_selection_service.dart
@@ -319,14 +319,17 @@ class _MobileSelectionServiceWidgetState
     if (end != null) {
       if (dragMode == MobileSelectionDragMode.leftSelectionHandler) {
         updateSelection(
-          selection.copyWith(
+          Selection(
+            start: _panStartSelection!.normalized.end,
             end: end,
-            start: _panStartSelection?.normalized.end,
-          ),
+          ).normalized,
         );
       } else if (dragMode == MobileSelectionDragMode.rightSelectionHandler) {
         updateSelection(
-          selection.copyWith(end: end),
+          Selection(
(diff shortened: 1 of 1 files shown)


lib/src/editor/editor_component/service/selection/mobile_selection_service.dart
```diff
@@ -318,15 +319,22 @@ class _MobileSelectionServiceWidgetState
...
     if (end != null) {
       if (dragMode == MobileSelectionDragMode.leftSelectionHandler) {
+        final positions = [end, _panStartSelection?.normalized.end];
+        positions.sort();
         updateSelection(
           selection.copyWith(
-            end: end,
-            start: _panStartSelection?.normalized.end,
+            end: positions.last,
+            start: positions.first,
```
LucasXu0, line 327: While both the start and end have been assigned, it would be better to create a new selection and normalize it.

LucasXu0:
```dart
    if (_panStartOffset == null ||
        _panStartScrollDy == null ||
        _panStartSelection == null) {
      return;
    }
// ...
      if (dragMode == MobileSelectionDragMode.leftSelectionHandler) {
        updateSelection(
          Selection(
            start: _panStartSelection!.normalized.end,
            end: end,
          ).normalized,
        );
      } else if (dragMode == MobileSelectionDragMode.rightSelectionHandler) {
        updateSelection(
          Selection(
            start: _panStartSelection!.normalized.start,
            end: end,
          ).normalized,
        );
      }
```

LucasXu0:
@dggbond Please remove the changes in `path.dart` and `position.dart` if the above suggestion is accepted.

dggbond:
@LucasXu0 thanks for sugestion, it is more clearer than my code.

Issue #500, opened 2023-09-25
[Bug] the text within the <mark> tag didn't parse correctly when pasting HTML.


LucasXu0:
### source
<img width="716" alt="Screenshot 2023-09-25 at 14 44 21" src="https://github.com/AppFlowy-IO/appflowy-editor/assets/11863087/6a3f2325-aded-4a49-9e83-522032f79596">

### editor
<img width="748" alt="Screenshot 2023-09-25 at 14 44 04" src="https://github.com/AppFlowy-IO/appflowy-editor/assets/11863087/ab30a135-983e-4d46-96f6-d3585c3ca2e5">

the texts with <mark> tag were missed, like the **dynamic tracking**.

### How to Reproduce

paste this html content.

```
[...]

PR #501, opened 2023-09-25
fix: the text within the <mark> tag didn't parse correctly when pasting HTML

LucasXu0:
### source
Screenshot 2023-09-25 at 14 44 21

### now
Screenshot 2023-09-25 at 15 07 54


LucasXu0:
closes https://github.com/AppFlowy-IO/appflowy-editor/issues/500

Issue #372, opened 2023-08-13
[Bug] Checkbox unresponsive in custom Theme example in editor example app


Mukund-Tandon:
### Bug Description

![ezgif com-video-to-gif (4)](https://github.com/AppFlowy-IO/appflowy-editor/assets/71614009/feb8191a-64a7-4081-82b5-53a807220c57)

In custom theme example of the editor , the checkbox is unresponsive to clicks

### How to Reproduce

- Run the example app of AppFlowy Editor 
- Go to side navigation bar and select custom color theme
- try to click on checkbox

### Expected Behavior

The checkbox should be responsive to clicks

### Operating System

MacOs

### AppFlowy Editor Version(s)

1.2.3

LucasXu0:
Closed as completed.

PR #543, opened 2023-10-18
feat: Checkbox unresponsive in custom Theme example in editor example app

johansutrisno:
linked to [#372](https://github.com/AppFlowy-IO/appflowy-editor/issues/372)

[Bug] Checkbox unresponsive in custom Theme example in editor example app

## Before

https://github.com/AppFlowy-IO/appflowy-editor/assets/22239637/48fdfbbb-16b9-4074-bbf0-b2db33e1606c

## After

https://github.com/AppFlowy-IO/appflowy-editor/assets/22239637/2de267fc-80d9-4d03-be29-c3299e5afa9c


example/lib/pages/customize_theme_for_editor.dart
```diff
@@ -102,10 +102,20 @@ class _CustomizeThemeForEditorState extends State<CustomizeThemeForEditor> {
...
-            size: 20,
-            color: Colors.white,
+          return InkWell(
+            onTap: () async {
+              final state = await editorState;
+              final transaction = state.transaction
+                ..updateNode(node, {
+                  TodoListBlockKeys.checked: !checked,
+                });
+              state.apply(transaction);
```
MayurSMahajan, line 112: what if we directly did `await editorState.apply(transaction)` instead of creating another variable. I am asking if it would make any difference? @Xazin @LucasXu0

MayurSMahajan (approved):

example/lib/pages/customize_theme_for_editor.dart
johansutrisno: Hi @MayurSMahajan thanks for your feedback.

I think it would be better to create a state variable, since we use it twice to get the transaction and to apply its state. Or do you have any concern?

MayurSMahajan: Yeah @johansutrisno I agree!

MayurSMahajan: Yeah I agree @johansutrisno !

Xazin: EditorState is already resolved at this point, change the method `customBuilder()` to
```
  /// customize the block style
  Map<String, BlockComponentBuilder> customBuilder(
    EditorState editorState,
  ) {
```
And then maybe something like this instead.
```
            onTap: () => editorState.apply(
              editorState.transaction
                ..updateNode(node, {TodoListBlockKeys.checked: !checked}),
            ),
```

johansutrisno: I have resolved the issue. Thank you for your feedback. @Xazin

example/lib/pages/customize_theme_for_editor.dart
```diff
@@ -102,10 +104,16 @@ class _CustomizeThemeForEditorState extends State<CustomizeThemeForEditor> {
         configuration: configuration,
         iconBuilder: (context, node) {
           final checked = node.attributes[TodoListBlockKeys.checked] as bool;
-          return Icon(
-            checked ? Icons.check_box : Icons.check_box_outline_blank,
-            size: 20,
-            color: Colors.white,
+          return InkWell(
```
Xazin, line 107: I also had another thought, since we do not need the splashFactory (the ink effect), maybe a GestureDetector will do?

Xazin (approved):

example/lib/pages/customize_theme_for_editor.dart
johansutrisno: Okay, I have switched from InkWell to GestureDetector. @Xazin

PR #553, opened 2023-10-25
feat: power checkbox shortcut

Xazin:
Add an easy way to add LogicalKeyboardKey triggers to the Todo Block, to power toggle a checkbox, which toggles all of it's children to the same state of the todo that is toggled.

https://github.com/AppFlowy-IO/appflowy-editor/assets/42929161/da5b9854-d023-4912-96de-ed97e3305e9a


Xazin:
Relates: https://github.com/AppFlowy-IO/AppFlowy/issues/3258

Issue #558, opened 2023-10-27
[Bug] When using Editor in new projects a Range Error is thrown if no header is passed but a footer is passed.


MayurSMahajan:
### Bug Description

When using AppFlowyEditor as a package in my Codeblock project. I noticed that a range error occurs when I initialize my Editor as follows:

```dart
final editorState = EditorState(
  document: Document(root: pageNode(children: [codeBlockNode()])),
);
//...
Widget build(BuildContext context) {
    return AppFlowyEditor(
      editorState: editorState,
      autoFocus: true,
      characterShortcutEvents: characterShortcutEvents,
      commandShortcutEvents: commandShortcutEvents,
      blockComponentBuilders: blockComponentBuilders,
      footer: const SizedBox(height: 200),
    );
  }
```

### How to Reproduce

I am not sure how one can reproduce this error.

1. Initialize an AppFlowyEditor instance with only a footer and no header.
2. Try running the project.

### Expected Behavior

No range error should occur 

### Operating System

Windows11

### AppFlowy Editor Version(s)

1.4.4

### Screenshots

![image](https://github.com/AppFlowy-IO/appflowy-editor/assets/47064215/8f3af6c4-6e9c-4b31-96e8-a944dda0c8c4)

PR #559, opened 2023-10-27
fix: check if header exists for footer

MayurSMahajan:
Solves #558

### DIFF
--- a/lib/src/editor/editor_component/entry/page_block_component.dart
+++ b/lib/src/editor/editor_component/entry/page_block_component.dart
@@ -93,7 +93,8 @@ class PageBlockComponent extends BlockComponentStatelessWidget {
               child: header!,
             );
           }
-          if (footer != null && index == items.length + 1) {
+          if (footer != null &&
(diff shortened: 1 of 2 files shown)


lib/src/editor/editor_component/entry/page_block_component.dart
```diff
@@ -93,7 +93,8 @@ class PageBlockComponent extends BlockComponentStatelessWidget {
               child: header!,
             );
           }
-          if (footer != null && index == items.length + 1) {
+          if (footer != null &&
+              index == items.length + (header != null ? 1 : 0)) {
```
Xazin, line 97: Will this do?
```suggestion
          if (footer != null && index == items.length + extentCount) {
```

Xazin: `extentCount` is 0 if no footer and no header, 1 if either, and 2 if both.

LucasXu0:
@MayurSMahajan please include the tests.

1. Header is null, but the footer is not.
2. Footer is null, but the header is not.

PR #567, opened 2023-10-30
changed padding values and added borderRadius

Kritarthsharma:
### Fixes [https://github.com/AppFlowy-IO/AppFlowy/issues/3313](url)

- Modified the appflowy_editor package to implement the requested UI fix.
- Changed the padding values and added borderRadius for more consistent look.

https://github.com/AppFlowy-IO/appflowy-editor/assets/42487588/15c8e25b-1bce-4375-b8d1-57ddd8516ea9

Screenshot 2023-10-30 at 6 00 11 PM
Screenshot 2023-10-30 at 5 59 42 PM


MayurSMahajan:
Hey @Kritarthsharma, 

actually, you were not supposed to add a border radius on the option, but the container which displays the options. Basically when there are no results the container has less border radius. 

image

Also, I see that you are padding to the no-results text. Rather than applying padding to the No Results text, you can calculate the width of other options used in the Selection Menu and then use that fixed width with center-aligned text. See if that is possible.

Kritarthsharma:
Hi @MayurSMahajan, I've made changes as you suggested. Can you please review it?

Issue #574, opened 2023-11-06
[Bug] PageStorage.of() context does not contain PageStorage widget


MichalNemec:
### Bug Description

Upon clicking in the editor i get this:
PageStorage.of() was called with a context that does not contain a pageStorage widget.
No PageStorage widget ancestor could be found starting from the context that was passed to PageStorage.of(). This can happen because you are using a widget that looks for a PageStorage ancestor, but no such ancestor exists.
The context used was:
ScrollablePositionedList(dirty, state: _ScrollablePositionedListState#d8f83(lifecycle state: created)

### How to Reproduce

click in the editor

### Expected Behavior

no error.

### Operating System

MacOS

### AppFlowy Editor Version(s)

1.4.4

### Screenshots

![image](https://github.com/AppFlowy-IO/appflowy-editor/assets/5663019/7538130a-92ce-41d7-8701-4db672a83e5b)

### Additional Context

This has been working last month, the problem is that i had in pubspec:
```
appflowy_editor:
    git:
      url: https://github.com/AppFlowy-IO/appflowy-editor.git
```
and then i updated to 1.4.4 and fixed implementation, there was an issue with ToolbarItem builder props difference.
i have editor in overlay:
```
overlayEntry = OverlayEntry(
      builder: (BuildContext _) {
        EditorState? localEditorState;
        return SafeArea(
          child: Scaffold(
            appBar: AppBar(
              leading: IconButton(
                icon: const Icon(Outline.close),
                onPressed: () {
                  FocusManager.instance.primaryFocus?.unfocus();
                  _removeFullscreenOverlay();
                },
              ),
              actions: [
                IconButton(
                  icon: const Icon(Outline.tick),
                  onPressed: () {
                    _removeFullscreenOverlay();

                    setState(() {
                      _json = json.encode(localEditorState?.document.toJson());
                    });
                  },
                ),
                const SizedBox(
                  width: 8,
                ),
              ],
            ),
            body: Container(
              padding: const EdgeInsets.symmetric(horizontal: 16.0, vertical: 12.0),
              child: FlowyEditor(
                jsonString: _json,
                enableDefinitions: false,
                autofocus: true,
                editable: true,
                showExplanation: true,
                onEditorStateChange: (editorState) {
                  localEditorState = editorState;
                },
              ),
            ),
          ),
        );
      },
    );

    Overlay.of(context).insert(overlayEntry!);
```

LucasXu0:
[...]

PR #575, opened 2023-11-06
fix: 'PageStorage.of' returns null

LucasXu0:


LucasXu0:
closes https://github.com/AppFlowy-IO/appflowy-editor/issues/574

PR #584, opened 2023-11-09
fix:Solve the left and right key bug. 

q200892907:
Solve the left and right key bug. When adding an input box at the head or tail of the editor, it cannot be moved.

### DIFF
--- a/lib/src/editor/editor_component/service/keyboard_service_widget.dart
+++ b/lib/src/editor/editor_component/service/keyboard_service_widget.dart
@@ -26,7 +26,8 @@ class KeyboardServiceWidget extends StatefulWidget {
 }
 
 @visibleForTesting
-class KeyboardServiceWidgetState extends State<KeyboardServiceWidget> implements AppFlowyKeyboardService {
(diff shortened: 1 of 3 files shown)


Xazin (approved):
CC @LucasXu0

lib/src/editor/editor_component/service/keyboard_service_widget.dart
```diff
@@ -52,8 +51,7 @@ class KeyboardServiceWidgetState extends State<KeyboardServiceWidget>
         return true;
       },
     );
-    editorState.service.selectionService
-        .registerGestureInterceptor(interceptor);
+    editorState.service.selectionService.registerGestureInterceptor(interceptor);
```
Xazin, line 54: Seems like your formatting options are different than default 80 character line length.

Xazin (changes requested):

LucasXu0:
@q200892907 The format step failed. PTAL.

q200892907:
> @q200892907 The format step failed. PTAL.

Restored default format

PR #600, opened 2023-11-21
fix: selection menu positioning on right half of editor

zoli:
part of https://github.com/AppFlowy-IO/AppFlowy/issues/3414.

The problem was presupposing that the editor width equal to the application width. So when there was something like a sidebar in AppFlowy the calculations were going wrong.

There will be still issues with the selection menu because the width is too long. It should get shorter and other items should get access through scroll. 
Personally, I prefer a vertical menu rather than a horizontal one.


LucasXu0:
> Personally, I prefer a vertical menu rather than a horizontal one.

Agreed. We should redesign the slash menu on desktop platform.

Issue #382, opened 2023-08-14
[FR]: Adding an ability to have a link check before embedding


yatendra2001:
### Description

There should be a mechanism to check whether the inserting text is in a valid link format. I think we can use regex for link validation.

![gif](https://global.discourse-cdn.com/standard11/uploads/appflowy/original/1X/d5d27e1daf1a762a5b5b5859071c33470da6cded.gif)

## Ideal Behaviour

![ideal gif](https://global.discourse-cdn.com/standard11/uploads/appflowy/original/1X/af2cfa6cb153205787c6b95ec4ec8f6c4bc0c0cc.gif)

### Impact

Users will make fewer mistakes while embedding the link hence creating a reliable ecosystem.

johansutrisno:
Hi @LucasXu0 and @Xazin I have added the feature for validate the link, could you please assign me to this issue and review the PR? Let me know if you have any suggestion. Thank you very much.

LucasXu0:
closed as completed.

PR #603, opened 2023-11-23
feat: adding an ability to have a link check before embedding

johansutrisno:
Linked to https://github.com/AppFlowy-IO/appflowy-editor/issues/382

### Before

https://github.com/AppFlowy-IO/appflowy-editor/assets/22239637/d203319e-17dc-4fe7-91ed-a297e95b302b

### After

https://github.com/AppFlowy-IO/appflowy-editor/assets/22239637/2a6e414d-e92b-487a-8292-0df0b0ec03c8


lib/src/editor/toolbar/utils/url_validator.dart
```diff
@@ -0,0 +1,15 @@
+class UrlValidator {
```
LucasXu0, line 1: Is it equal to the `isURL` function?

johansutrisno: I see, yes, you're right. i will remove this class and replace with isURL function.

johansutrisno:
Hi @LucasXu0 , I have resolved your feedback, are there any concerns or other feedback? Thank you very much. 🙌

Issue #542, opened 2023-10-18
[Bug] focus issue


q200892907:
### Bug Description

header/footer is TextField，focus issus

### How to Reproduce

First click the editor, then click the header/footer TextField
My environment is flutter 3.13

### Expected Behavior

focus correct

### Operating System

desktop

### AppFlowy Editor Version(s)

master

### Screenshots

https://github.com/AppFlowy-IO/appflowy-editor/assets/4517301/08462521-b177-46e0-ab48-164d00f779e7

### Additional Context

#520 unsolved
flutter 3.13.x

LucasXu0:
I attempted to upgrade the minimum supported version to 3.13; however, I encountered too many unexpected issues. Therefore, I recommend using Flutter 3.10 instead for now.

Similar to this issue, it functions properly in Flutter 3.10.x but not in 3.13.x.

q200892907:
Left and right key operations are blocked in the input boxes at the head and tail of the editor. @LucasXu0

PR #607, opened 2023-11-28
fix: header/footer focus issue

q200892907:
https://github.com/AppFlowy-IO/appflowy-editor/assets/4517301/f095809d-c430-452d-99e9-eeac8fe54c3b

#542


q200892907:
@LucasXu0 Solve the problem of some shortcut operation keys failing when adding input boxes at the beginning or end, including select all cut undo

Issue #610, opened 2023-11-30
[Bug] The shortcut keys in the input method are blocked by the editor


q200892907:
### Bug Description

The shortcut keys in the input method are blocked by the editor

### How to Reproduce

For example, when using the system Chinese input method, the operation of switching word selection using the up, down, left and right keys is no longer available.

https://github.com/AppFlowy-IO/appflowy-editor/assets/4517301/93cf634c-0727-4e0d-8842-5028623668b5

### Expected Behavior

When using the input method, do not display its shortcut keys and make the cursor position operable.

https://github.com/AppFlowy-IO/appflowy-editor/assets/4517301/10393bdb-4f7e-42bb-af81-c3dddf73c8ad

### Operating System

desktop

### AppFlowy Editor Version(s)

main

q200892907:
@LucasXu0 @Xazin

annieappflowy:
Hi @q200892907 , thanks for your active participation in our projects!
We're eager to gather valuable feedback from our users to better plan for 2024. Participants who complete this survey will gain priority access to our new launches. We appreciate your insights: https://tally.so/r/nW8qAQ

PR #652, opened 2024-01-06
fix:chinese input 610

q200892907:
#610  test window/linux @LucasXu0


LucasXu0:
Have you tested it on Windows?

q200892907:
> Have you tested it on Windows?


I have tested on Windows and it can be used, but the test is not comprehensive enough.

q200892907:
I don’t know what problem the Windows judgment added at that time was used to solve. You can do a test based on the problem at that time.

q200892907:
https://github.com/AppFlowy-IO/appflowy-editor/assets/4517301/73ca0858-565a-4358-8e04-91ac524fee8e

windows test video

LucasXu0:
@q200892907 Thanks for contributing.

PR #659, opened 2024-01-09
fix: support character deletion while holding down the Shift key.

Jayaprakash-dev:
This fix is related to the bug in [AppFlowy-IO/AppFlowy](https://github.com/AppFlowy-IO/AppFlowy):

> [Bug] Shift blocks backspace deletion # 4279
> [link]: https://github.com/AppFlowy-IO/AppFlowy/issues/4279

I've addressed the reported issue with the following changes:

**Issue:**
When the shift key is held down, deleting a character or selection by pressing backspace is not functioning as expected.

**Solution:**
I've implemented a fix by assigning the same Backspace handler in `backspace_command.dart` to another command shortcut event handler named `deleteLeftCharacterCommand` which is responsible for handling `shift+backspace` shortcut event. This adjustment ensures that the deletion functionality works seamlessly, resolving the reported bug.

**Repository Context:**
Although the issue is posted in another related repository **AppFlowy-IO/AppFlowy**, the code files relevant to this bug are present in this repository. Therefore, I've created this PR here.

**Additional Changes:**
I've also included the required test files to ensure the reliability of this fix.


LucasXu0:
It's not necessary to define a new shortcut command for Shift+Backspace. Simply adding this shortcut (Shift+Backspace) to the existing Backspace command is enough.

Jayaprakash-dev:
**@LucasXu0**, Just got to know, that new shortcuts can be added to existing command handlers through a comma-separated string. As a result, I removed the unnecessary `deleteLeftCharacterCommand` handler along with its unit tests. However, I retained the widget tests for the `shift+backspace` command event.

PR #665, opened 2024-01-10
feat: support formatting greater hyphen to single arrow

Jayaprakash-dev:
This PR is related to feature request made in [AppFlowy-IO/AppFlowy](https://github.com/AppFlowy-IO/AppFlowy) repo

[issue]: [[FR] - + > then → #4277](https://github.com/AppFlowy-IO/AppFlowy/issues/4277)

Added a feature that enable users to type - + > to insert → in the editor


Xazin (approved):

Xazin:
The CI fails, can you format the files with the dart formatter?

We use 80 character line length, if you have upped it in your local environment you might have to change it back to 80 and then format the files.

Jayaprakash-dev:
I've formatted the code file using dart SDK formatter.

lib/src/editor/block_component/standard_block_components.dart
```diff
@@ -97,6 +97,9 @@ final List<CharacterShortcutEvent> standardCharacterShortcutEvents = [
 
   // convert => to arrow
   formatGreaterEqual,
+
+  // convert -> to single arrow (→)
+  formatGreaterHyphen,
```
LucasXu0, line 102: IMO, it should not be added to the standard shortcuts as it is not a commonly used shortcut.

Jayaprakash-dev: I thought it was fine to include `formatGreaterHyphen` in the `standardCharacterShortcutEvents` list since it already includes `formatGreaterEqual`. However, if you suggest not adding formatGreaterHyphen there, an alternative approach would be to include it in the `characterShortcutEvents` list within the `AppFlowyEditorPage` class located in [**AppFlowy-IO/AppFlowy**](https://github.com/AppFlowy-IO/AppFlowy/blob/main/frontend/appflowy_flutter/lib/plugins/document/presentation/editor_page.dart) repo.

If you're okay with that, then I could close the current pull request and submit a new one in **AppFlowy** repo.

Xazin: Including it in the editor is fine, you can remove it from the standard shortcuts, then in AppFlowy we can selectively pick it if need be.

Jayaprakash-dev: @**Xazin**, sure I'll commit those changes, btw there's currently no test case for formatting greater equal to double arrow `= + > to insert ⇒`. If you'd like, I can also create a test case for that scenario.

annieappflowy: auto format -> 
cmd z to cancel

Issue #658, opened 2024-01-09
[Bug] Selecting heading, list or quote opens keyboard, but menu is not dismissed


armandsLa:
### Bug Description

When selecting heading, list or quote option in menu, the keyboard gets opened, but the menu is not dismissed. This causes an issue the next time toolbar item is selected, because user will have to press it twice. First press will close the still open menu.

This happens only when there is no text selection. If something is selected then the keyboard stays hidden.

### How to Reproduce

1. Open editor
2. Press toolbar item to get menu for headings
3. Select "Heading 1"
4. Try selecting the same toolbar icon again

### Expected Behavior

Menu to be always closed when keyboard opens

### Operating System

iOS

### AppFlowy Editor Version(s)

2.2.0

PR #670, opened 2024-01-11
fix: the menu dismissed after selecting block style

LucasXu0:


LucasXu0:
closes https://github.com/AppFlowy-IO/appflowy-editor/issues/658

PR #680, opened 2024-01-19
Localize more items

jpenna:
This is related to the PR https://github.com/AppFlowy-IO/AppFlowy/pull/4426

It should be merged before the one in the app is merged. 

---

I'm also doing the PT-BR translation in this PR to verify all texts are correctly replaced.


LucasXu0:
Hi, @jpenna. I converted it to a draft PR. Please mark it as ready for review if it's ready.

lib/src/editor/block_component/table_block_component/table_block_component.dart
```diff
@@ -251,7 +251,7 @@ class _TableBlockComponentWidgetState extends State<TableBlockComponentWidget>
 }
 
 SelectionMenuItem tableMenuItem = SelectionMenuItem(
-  name: AppFlowyEditorL10n.current.table,
+  getName: () => AppFlowyEditorL10n.current.table,
```
jpenna, line 254: A better way to achieve this same thing would be to rebuild everything if `Intil` changes, but then it would require a long refactor of many classes to use `Localizations.of(context)` or add some listener to rebuild everything. Probably something to be thought about later?

jpenna:
I think this one can be merged so it doesn't grow too big.
I can work on another PR with the translations used in the app related to the PR mentioned at the beginning).

jpenna:
I'll send another PR with the translations of commands. It will be quite large, so we can merge this one first.

LucasXu0:
@jpenna One test failed. Please check it.

LucasXu0 (approved):

PR #683, opened 2024-01-22
upgraded markdown component

GQAdonis:
I updated the version of the markdown component supported to 7.2.1 from 7.1.0 and fixed the issue where the NodeVisitor mixin was changed to an abstract class.  Also updated a variety of other component versions.


Xazin:
Closing this due to CLA not being signed, and unrelated changes to pubspec.yaml.

PR #687, opened 2024-01-24
Translate commands

jpenna:
Merge this one first https://github.com/AppFlowy-IO/appflowy-editor/pull/680

---

This PR adds translations for all commands.


lib/src/editor/editor_component/service/shortcuts/command_shortcut_event.dart
```diff
@@ -11,10 +11,11 @@ class CommandShortcutEvent {
     required this.key,
     required this.command,
     required this.handler,
+    required String Function()? getDescription,
```
jpenna, line 14: The value is optional for cases where we don't want to allow the user to override, for example, when pasting a link and converting that to a link.

LucasXu0:
@jpenna I have merged your previous PR already. Please resolve the conflicts.

Issue #688, opened 2024-01-24
[Bug] Pasting unformatted text in formatted text does not format


jpenna:
### Bug Description

Pasting pastes with format even when there is none (default background, for example).

Pasting without format also changes the format at the destination.

### How to Reproduce

1. Add color to the background of a text
2. Copy some text
3. Paste it between the formatted text as unformatted text (`ctrl+shift+v`)
4. See the pasted text does not have background

### Expected Behavior

~The copied text should be pasted keeping the format of the destination area when it is an unformatted text.~

~At least,~ using `ctrl+shift+v` should paste plain text and keep the destination format, but it doesn't. 

### Operating System

Linux

### AppFlowy Editor Version(s)

Built from source

### Screenshots

[Screencast from 24-01-2024 11:03:01.webm](https://github.com/AppFlowy-IO/appflowy-editor/assets/16005946/3ea13b6d-706c-4946-9907-a673d7731ef2)

LucasXu0:
@jpenna You were pasting the text without formatting, but you expected the text to follow the style of the destination text, correct?

jpenna:
Yes, that's it. I found a solucionar already. I'll send a PR today
probably. Working on tests.

jpenna:
Sorry, I replied on the run and didn't realize it was another issue I've opened. I was working on another fix and not this one.

I'll work on this tomorrow.

PR #692, opened 2024-01-26
Use format of wrapping node when pasting plain text

jpenna:
closes https://github.com/AppFlowy-IO/appflowy-editor/issues/688


test/new/service/shortcuts/command_shortcut_events/paste_command_test.dart
jpenna, line 1: Shortcut's test folder isn't following the structure of `src`. Should this be refactored?

jpenna:
Could you run the checks again, please? I think they failed because I force pushed while they were running.

Issue #696, opened 2024-01-29
[Bug] Input characters are out of order when using some IME


abcamus:
### Bug Description

When I am using sogou input engine to type in Chinese charactors，the words are out of order.

### How to Reproduce

using sogou input to enter “今天”，“一天”，“喝茶”，the sentence will be "space+今天一喝茶天"

https://github.com/AppFlowy-IO/appflowy-editor/assets/17241681/a4245493-d5d1-4627-b1d2-4532b8a43b0f

### Expected Behavior

input words in correct order

### Operating System

Windows 11

### AppFlowy Editor Version(s)

0.4.3

abcamus:
yeah,I’ll try to fix it on windows.

PR #699, opened 2024-01-30
fix: debounce of desktop IME (#696)

abcamus:
debounce for some IME, such as sogou Chinese input method.

### DIFF
--- a/lib/src/editor/editor_component/service/ime/non_delta_input_service.dart
+++ b/lib/src/editor/editor_component/service/ime/non_delta_input_service.dart
@@ -102,7 +102,7 @@ class NonDeltaTextInputService extends TextInputService with TextInputClient {
       debounceKey,
       PlatformExtension.isMobile
           ? const Duration(milliseconds: 10)
-          : Duration.zero,
+          : const Duration(microseconds: 1),
       () {
         currentTextEditingValue = value;
         apply(deltas);


LucasXu0:
@abcamus Why do we need to debounce it?

abcamus:
> @abcamus Why do we need to debounce it?

Hi, @LucasXu0, after searching the debug log, it is the Sogou input method that triggers updateEditingValue event multiple times. For example, when I type in '今'，the input value of updateEditingValue method is as follows:

1. an empty and composing, with offset = 1, composing = TextRange(start: 1, end: 1)
2. a composing value, with offset = 1, composing = TextRange(start: 1, end: 2)
3. final value, with offset = 2, composing = TextRange(start: -1, end: -1)

appflowy editor's current implementation will handle all of whese three events: applying the delta -> notify editor's selection -> try to attach textInput

this procedure seems redundant and buggy when using Sogou input method.

My suggestion is just try to apply delta only when composing is empty, like this:
```dart
@override
  void updateEditingValue(TextEditingValue value) {
    if (currentTextEditingValue == value || value.composing != TextRange.empty) {
      return;
    }
```

It works fine on windows desktop, while I have no idea on any other side effects. Also the previous debounce change just luckily avoids the buggy procedure.

LucasXu0:
Thanks for the detailed info. I will look into your suggestion.

PR #700, opened 2024-01-31
fix: fix checkbox not work as expected in html module.

penkzhou:
First, let me describe how the bug was discovered. I was using this lib to export Markdown to PDF. The library first renders the Markdown in an editor on the page by calling the `markdownToDocument` method. It then converts the rendered document to HTML code via the `documentToHtml` method. Finally, it exports the generated HTML into a PDF file.

To visualize the steps, here are some screenshots:

| editor ui    | bug  export to pdf |  normal export to pdf |
| -------- | ------- |  ------- |
|   |    |    |

### bug cause
according the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox), checkbox should write like this:
#### checked checkbox
`<input type="checkbox" id="scales" name="scales" checked />
    <label for="scales">Scales</label>`

#### non check checkbox(without `checked` attribute)
`<input type="checkbox" id="scales" name="scales" />
    <label for="scales">Scales</label>`

### how to fix it
change the code in the `todo_list_node_parser.dart` file. main change include following points:

1. change the insert position of `input` tag to keep checkbox in the first of the line.
2. change the way of setting `checkbox`'s `checked` attribute.


LucasXu0:
LGTM.

PR #705, opened 2024-02-06
feat: word counter service

Xazin:
My first idea was to setup a listener to the Transaction stream of the EditorState, and then update the counters only with the changeset, but this seems not quite possible with how operations currently work.

Currently, an UpdateOperation contains inserts in the delta, and whilst we do have the old attributes to compare against, it seems that it can happen that the new attributes delta can be a subset of the old one, causing it to be near impossible to rely on the difference in characters and words as the comparison of the two might not be representative of the actual changes to the document.

We can improve and build on this (eg. add Selection word/char count to the service) in the future.

### DIFF
--- a/lib/src/plugins/word_count/word_counter_service.dart
+++ b/lib/src/plugins/word_count/word_counter_service.dart
@@ -20,11 +20,9 @@ final _wordRegex = RegExp(r"\w+(\'\w+)?");
 /// the counter stats.
 ///
 class WordCountService with ChangeNotifier {
-  WordCountService({
-    required EditorState editorState,
-  }) : _editorState = editorState;
+  WordCountService({required this.editorState});
 
-  final EditorState _editorState;
+  final EditorState editorState;
 
   int _wordCount = 0;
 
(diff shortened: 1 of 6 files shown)


lib/src/plugins/word_count/word_counter_service.dart
```diff
@@ -0,0 +1,129 @@
...
+/// method, this will add a listener to the [Transaction] updates
+/// of the [EditorState], and do an initial run-through to populate
+/// the counter stats.
+///
+class WordCountService with ChangeNotifier {
+  WordCountService({
+    required EditorState editorState,
+  }) : _editorState = editorState;
+
+  final EditorState _editorState;
```
LucasXu0, line 27: is it necessary to mark it as private? if not, then using required this.editorState seems better.

lib/src/plugins/word_count/word_counter_service.dart
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  (int, int) _countersFromNode(Node node) {
+    int wCount = 0;
+    int cCount = 0;
+
+    final plain = _toPlainText(node);
+    wCount += _wordsInString(plain);
+    // TODO: Account for eomjis having length > 1
+    cCount += plain.length;
```
LucasXu0, line 115:
```suggestion
    cCount += plain.runes.length
```

LucasXu0: this will return the real length of the emojis.

Xazin: Thanks, I was looking for this

Issue #545, opened 2023-10-20
[Bug]  Flutter web Slash command menu Overflows


dawarepramod4:
### Bug Description

When i use margin to any parent widget in the widget tree the / command menu goes to out of the screen.
you can see in the below image the slash command menu is out of the screen. sometimes it goes all out.
![image](https://github.com/AppFlowy-IO/appflowy-editor/assets/91264300/0b76e1d7-efa8-4dea-aeed-05a95ff96dfd)

### How to Reproduce

In the example code 
1. Remove the padding from "example/lib/pages/desktop_editor.dart" 

![Screenshot from 2023-10-20 21-36-11](https://github.com/AppFlowy-IO/appflowy-editor/assets/91264300/2bb816a7-a1ca-4a17-9f7a-d53f92447185)

3. Add below line in the "example/lib/pages/editor.dart" file container.
margin: const EdgeInsets.symmetric(horizontal: 200),
![image](https://github.com/AppFlowy-IO/appflowy-editor/assets/91264300/4cc321c7-e8ff-4229-af62-72203635afe7)

### Expected Behavior

The slash Command Menu should not overflow from screen and should be adjusted in the visible area of the widget.

### Operating System

Ubuntu 23.04

### AppFlowy Editor Version(s)

1.4.4 

### Screenshots

-

### Additional Context

-

dawarepramod4:
Or If i give width to the parent container of the editor then also the position mistmatches

stevenosse:
Hi @LucasXu0 
I think this issue can be closed

LucasXu0:
closed as completed.

PR #709, opened 2024-02-08
fix: Slash command menu bad positioning in Web Version

dagyu:
## Description:
While developing with AppFlowy editor, I encountered the following bug in the web version. I have integrated the AppFlowy editor into my app as shown in the image. On the left side, the editor is positioned between a sidebar and a header. This placement causes a displacement of the command menu. The slippage occurs because the menu is inserted relative to its container, rather than being positioned absolutely within the web page.

Screenshot 2024-02-08 at 16 04 57

## Expected Behavior:
The command menu should be positioned correctly without any slippage, even when the editor is placed between a sidebar and a header.

## Fix:
To resolve this issue, I modified Overlay.of(context) to Overlay.of(context, rootOverlay: true). Additionally, this fix also addresses issue #545.

Please review and consider implementing this fix in the next release. Let me know if you need any further clarification or information. Thank you!

### DIFF
--- a/lib/src/editor/block_component/table_block_component/table_action_menu.dart
+++ b/lib/src/editor/block_component/table_block_component/table_action_menu.dart
@@ -122,7 +122,7 @@ void showActionMenu(
       );
     },
   ).build();
-  Overlay.of(context).insert(overlay!);
+  Overlay.of(context, rootOverlay: true).insert(overlay!);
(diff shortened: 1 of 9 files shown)


LucasXu0:
@dagyu Thanks for your contribution. LGTM.

PR #713, opened 2024-02-12
Fix md image export

rileyhawk1417:
When exporting the document to markdown the image would show up as ``. The image url/src didn't get passed.

In the [image_node_parser.dart](https://github.com/AppFlowy-IO/appflowy-editor/blob/e09bde476bf33712cb59d11949b22981b27a3eb2/lib/src/plugins/markdown/encoder/parser/image_node_parser.dart#L9C1-L12C4) its looking for the attribute `image_src` which doesnt exist.
```
  @override
  String transform(Node node, DocumentMarkdownEncoder? encoder) {
    return '![](${node.attributes['image_src']})';
  }
```

However when checking the attributes `url` is passed and not `image_src`


lib/src/plugins/markdown/encoder/parser/image_node_parser.dart
```diff
@@ -8,6 +8,6 @@ class ImageNodeParser extends NodeParser {
 
   @override
   String transform(Node node, DocumentMarkdownEncoder? encoder) {
-    return '![](${node.attributes['image_src']})';
+    return '![](${node.attributes['url']})';
```
LucasXu0, line 11:
```suggestion
    return '![](${node.attributes[ImageBlockKeys.url]})';
```

LucasXu0:
@rileyhawk1417 This test failed.

```dart

    test('parser image node', () {
      final node = Node(
        type: 'image',
        attributes: {
          'image_src': 'https://appflowy.io',
        },
      );

      final result = const ImageNodeParser().transform(node, null);
      expect(result, '![](https://appflowy.io)');
    });
```

lib/src/plugins/markdown/encoder/parser/image_node_parser.dart
rileyhawk1417: So more like
`ImageBlockKeys.url='url'` ?

LucasXu0: Yes. 

Screenshot 2024-02-18 at 10 58 12

LucasXu0: I committed the suggestion.

Issue #693, opened 2024-01-27
[Bug] Pasting text in plain text with URL makes everything a link


jpenna:
### Bug Description

Pasting a text in plain text that contains a URL turns the whole text in the URL and formats everything as a link.

### How to Reproduce

1. Copy a text that contains a URL in the format `http://...`
2. Paste in plain text (`ctrl+shift+v`)
3. See everything is turned into the URL

### Expected Behavior

Should only convert the URL part to the link.

### Operating System

Linux

### AppFlowy Editor Version(s)

Built from source

### Screenshots

[Screencast from 27-01-2024 08:16:20.webm](https://github.com/AppFlowy-IO/appflowy-editor/assets/16005946/d71f1c3e-7116-4801-8fe4-a81b10e6c8b9)

### Additional Context

Offending code is in https://github.com/jpenna/appflowy-editor/blob/688-plain_text/lib/src/editor/editor_component/service/shortcuts/command/paste_command.dart#L104

PR #737, opened 2024-03-05
fix: transfer the whole line into link when text contains url (#693)

penkzhou:
### bug
https://github.com/AppFlowy-IO/appflowy-editor/issues/693

### fix approach

1. detect the text with url regrex
2. if there is a url exist, split the text into three parts. The first part and the final part may not exist.add url attribute for the url part.
3. if there is no url exist,just return the text.

### test status
unit test: all pass.


jpenna:
Looks good! Worth adding tests to the highlighted lines in the "Files changed" tab.

jpenna:
Great, thanks! You have to fix the commit message, but otherwise it looks good to go

LucasXu0:
Never mind. I can squash this PR. LGTM.

Issue #589, opened 2023-11-15
[Bug] Up and down arrow problem


q200892907:
### Bug Description

Appears after setting a relatively large block spacing

flutter 3.13.9
appflowy-editor The latest version

### How to Reproduce

Watch video example

### Expected Behavior

normal movement

### Operating System

desktop

### AppFlowy Editor Version(s)

main

### Screenshots

![image](https://github.com/AppFlowy-IO/appflowy-editor/assets/4517301/35486d8c-f884-4c1e-a42e-ad996b0c88f8)

https://github.com/AppFlowy-IO/appflowy-editor/assets/4517301/21894a1a-e494-4862-bd95-25beecf5e2f9

Xazin:
I'm not sure what you mean by vertical padding affecting the movement you're doing.

You're moving within the same paragraph, can you elaborate on which part of the video showcases the issue?

q200892907:
I can't move to the next block if I keep pressing the arrow. Likewise, I can't move if I press the up arrow. It has no effect on the same block. @Xazin

Xazin:
On MacOS the up arrow works, but not the down arrow :+1:

PR #751, opened 2024-03-21
fix: incorrect cursor movement behaviour

### LINKED ISSUE #656: [Bug] Heading level specific padding stops cursor
### Bug Description

Adding a different padding for each heading level stops the cursor from moving up using the arrow key when it is at a heading.

### How to Reproduce

To reproduce, specify different padding for each heading level in the HeadingBlockComponentBuilder.

```dart
BlockComponentBuilder headingBlock(BuildContext context) {
  final baseHeadingStyle = TextStyle(
    letterSpacing: -1,
    fontWeight: FontWeight.w500,
    color: context.palette.primary,
  );

  return HeadingBlockComponentBuilder(
    configuration: BlockComponentConfiguration(
      padding: (node) {
        // No top padding for first line/paragraph
        if (node.previous == null) {
          return const EdgeInsets.only(top: 0, bottom: 8);
        }

        switch (node.attributes[HeadingBlockKeys.level]) {
          case 1:
            return const EdgeInsets.only(top: 20, bottom: 8);
          case 2:
            return const EdgeInsets.only(top: 17, bottom: 8);
          case 3:
            return const EdgeInsets.only(top: 14, bottom: 8);
          case 4:
            return const EdgeInsets.only(top: 11, bottom: 8);
          case 5:
            return const EdgeInsets.only(top: 8, bottom: 8);
          case 6:
            return const EdgeInsets.only(top: 5, bottom: 8);
          default:
            return const EdgeInsets.only(top: 5, bottom: 8);
        }
      },
    ),
    textStyleBuilder: (level) {
      switch (level) {
        case 1:
          return baseHeadingStyle.copyWith(fontSize: 22);
        case 2:
          return baseHeadingStyle.copyWith(fontSize: 19);
        case 3:
          return baseHeadingStyle.copyWith(fontSize: 17.7);
        case 4:
          return baseHeadingStyle.copyWith(fontSize: 16);
        case 5:
          return baseHeadingStyle.copyWith(fontSize: 14.5);
        case 6:
          return baseHeadingStyle.copyWith(fontSize: 13.8);
        default:
          return baseHeadingStyle.copyWith(fontSize: 13.8);
      }
    },
  );
}
```

### Expected Behavior

The cursor should be able to move up using the arrow key.

### Operating System

MacOS

2.2.0

### Screenshots

Screenshot 2024-01-09 at 5 46 23 AM

Unable to move my cursor to the 'Hi' line using the up arrow key.

### Additional Context

It works fine if I return only a single padding.

### LINKED ISSUE #745: [Bug] Wrong cursor up/down movement behaviour
### Bug Description

When you use the up/down arrow keys in a long line (which is split into multiple lines because of the screen width), the cursor moves up/down in the previous/next row instead of going up/down in the previous/next part of the current row.

### How to Reproduce

Open "AppFlowy 0.5.2" or the example app and create a line long enough to be split into several rows.
Press up/down keys.

### Expected Behavior

The cursors moves to the previous/next part of the multiline instead of go directly into previous/next line

### Operating System

MacOS 14.3.1 (Apple Silicon)

2.3.3 (latest)

### Screenshots

https://github.com/AppFlowy-IO/appflowy-editor/assets/46776108/984340ea-2ae8-4328-a8dc-9b2e8cf6d7a9

### Additional Context

In previous versions of AppFlowy this didn’t happen. Unfortunately I can’t say which one exactly

Jei-sKappa:
fixes #745 
Previously the `moveVertical` function only moved the cursor 'node to node' without taking into consideration multiline nodes or some portion of a line with different fontSize (ex: a paragraph with size 16 but with a bolded word with size 32). This was caused by the PR #657 that closes #656 and #589.

### DIFF
--- a/lib/src/extensions/position_extension.dart
+++ b/lib/src/extensions/position_extension.dart
@@ -69,29 +69,6 @@ extension PositionExtension on Position {
     EditorState editorState, {
     bool upwards = true,
   }) {
-    //* GET THE CURRENT OFFSET
-    final selection = editorState.selection;
-    final rects = editorState.selectionRects();
-    if (rects.isEmpty || selection == null) {
-      return null;
-    }
-
-    Offset currentOffset;
-    final Rect caretRect;
-    if (selection.isBackward) {
-      caretRect = rects.reduce(
-        (current, next) => current.bottom > next.bottom ? current : next,
-      );
-      currentOffset =
-          upwards ? throw Exception('Ivalid state') : caretRect.bottomRight;
-    } else {
-      caretRect = rects.reduce(
-        (current, next) => current.top <= next.top ? current : next,
-      );
-      currentOffset = upwards ? caretRect.topLeft : caretRect.bottomLeft;
-    }
-
-    //* GET THE CURRENT NODE'S TEXT HEIGHT
     final node = editorState.document.nodeAtPath(path);
     if (node == null) {
       return this;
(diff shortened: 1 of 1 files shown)


LucasXu0:
@Jei-sKappa Thanks for contributing.

Jei-sKappa:
@LucasXu0 Thanks for your reply! Based on the response of GitHub Action it seems like the problem is in the commit message and not in the code. Am I right?

LucasXu0:
> @LucasXu0 Thanks for your reply! Based on the response of GitHub Action it seems like the problem is in the commit message and not in the code. Am I right?

Except for the commit message, you need to format the Dart code too. Please click on the CI and check the error.

Jei-sKappa:
@LucasXu0 Sorry for the late response.
I just pushed a new update to the pull requests that should fix the test's errors in the CI.
Unfortunately my first commit text exceeds the length defined in commitlint and the CI keeps failing 'cause of this.
Can I ignore it? If not what should I do to fix this?

Xazin:
> Can I ignore it? If not what should I do to fix this?

You don't have to worry about the commitlint failing, it's not important :+1:

lib/src/extensions/position_extension.dart
```diff
@@ -67,66 +69,142 @@ extension PositionExtension on Position {
...
+
+    final nodeRenderBox = node.renderBox;
+    if (nodeRenderBox == null) {
+      return this;
+    }
+
+    final nodeSelectable = node.selectable;
+    if (nodeSelectable == null) {
+      return this;
+    }
```
Xazin, line 85: We can consolidate these three if-statements
```suggestion
    final node = editorState.document.nodeAtPath(path);
    final nodeRenderBox = node?.renderBox;
    final nodeSelectable = node?.selectable;
    if (node == null || nodeRenderBox == null || nodeSelectable == null) {
      return this;
    }
```

lib/src/extensions/position_extension.dart
```diff
@@ -67,66 +69,142 @@ extension PositionExtension on Position {
...
-      final rect = rects.reduce(
-        (current, next) => current.bottom >= next.bottom ? current : next,
+    // The offset of outermost part of the caret.
+    // Either the top if moving upwards, or the bottom if moving downwards.
+    final Offset caretOffset;
+
+    final Rect caretRect;
+    if (editorSelection.isBackward) {
+      caretRect = rects.reduce(
+        (current, next) => current.bottom > next.bottom ? current : next,
```
Xazin, line 100: Maybe something like this, at least for the `caretRect` I find these two versions better:
```
    final Rect caretRect = rects.reduce((current, next) =>
      editorSelection.isBackward 
          ? current.bottom > next.bottom ? current : next
          : current.top <= next.top ? current : next,
    );

    // The offset of outermost part of the caret.
    // Either the top if moving upwards, or the bottom if moving downwards.
    final Offset caretOffset =
        editorState.isBackward
            ? upwards ? caretRect.topRight : caretRect.bottomRight
            : upwards ? caretRect.topLeft : caretRect.bottomLeft;
```
---
```
    final Rect caretRect = rects.reduce((current, next) {
      if (editorSelection.isBackward) { 
          return current.bottom > next.bottom ? current : next;
      }
      return current.top <= next.top ? current : next;
    });

    // The offset of outermost part of the caret.
    // Either the top if moving upwards, or the bottom if moving downwards.
    final Offset caretOffset =
        editorState.isBackward
            ? upwards ? caretRect.topRight : caretRect.bottomRight
            : upwards ? caretRect.topLeft : caretRect.bottomLeft;
```

lib/src/extensions/position_extension.dart
```diff
@@ -67,66 +69,142 @@ extension PositionExtension on Position {
...
-    final position =
-        editorState.service.selectionService.getPositionInOffset(offset);
-
-    if (position != null && !position.path.equals(path)) {
-      return position;
+    final nodeConfig = editorState.service.rendererService
+        .blockComponentBuilder(node.type)
+        ?.configuration;
+    if (nodeConfig == null) {
+      // This should not happen.
```
Xazin, line 114:
```suggestion
      assert(nodeConfig != null, 'Block Configuration should not be null');
```

lib/src/extensions/position_extension.dart
```diff
@@ -67,66 +69,142 @@ extension PositionExtension on Position {
...
+    //   - If the caret is in the first/last line: this is the worst case
+    //      scenario, but only if the padding choosen by the user is very
+    //      large. (padding >= (multiline's textHeight - caretHeight) / 3
+    //      can start to be considered large. Note that in an average bad case
+    //      scenario the position will be found in 10/12 ms instead of 1/2 ms)
+    // - If the current node is not multiline: the cycle will be completely
+    //   skipped because `remainingMultilineHeight` would be 0.
+    Offset newOffset = caretOffset;
+    Position? newPosition;
+    for (var y = minFontSize;
```
Xazin, line 148:
```suggestion
    for (double y = minFontSize;
```

lib/src/extensions/position_extension.dart
```diff
@@ -67,66 +69,142 @@ extension PositionExtension on Position {
...
-        }
+    final padding = nodeConfig.padding(node);
+    final nodeRect = nodeSelectable.getBlockRect();
+    final nodeHeight = nodeRect.height;
+    final textHeight = nodeHeight - padding.vertical;
+    final caretHeight = caretRect.height;
+
+    // Minimum (acceptable) font size
+    // Consider augmenting this value to increase performance.
+    const minFontSize = 1.0;
```
Xazin, line 126:
```suggestion
    const double minFontSize = 1.0;
```

Xazin (commented):
We need to introduce some more tests so we can stop a regression from occurring.

We can add tests in a new PR, I will open a new issue once @LucasXu0 has taken a look at this PR.

Issue #755, opened 2024-04-03
[Bug]  Issue with Korean Input: Loss of TextEditingRange When Alternating Consonant and Vowel


kyungilcho:
### Bug Description

When typing in Korean, alternating between consonants and vowels, the text editor seems to lose `TextEditingRange` after the input of the first two characters. This results in incorrect character formation. Specifically, attempting to write a sequence of the same character "하" four times instead results in "하하ㅎㅏㅎㅏ".

Thank you for looking into this matter. I look forward to a resolution that will enhance the typing experience for Korean language users.

## Steps to Reproduce
1. Open the text editor.
2. Switch the input method to Korean.
3. Try typing "하하하하" by alternating input between consonants and vowels (i.e., type "ㅎ", then "ㅏ" four times in a sequence).
4. Observe that the input is incorrectly registered as "하하ㅎㅏㅎㅏ" instead of "하하하하".

## Expected Behavior
The editor should accurately reflect the Korean input without losing the `TextEditingRange`, allowing for the correct formation of characters. Typing "ㅎ" followed by "ㅏ" four times in succession should result in "하하하하".

## Actual Behavior
After typing the first two characters correctly as "하하", the editor loses track of the `TextEditingRange`. Subsequent inputs of "ㅎ" and "ㅏ" are not combined into "하", but are instead added separately, resulting in "하하ㅎㅏㅎㅏ".

### Operating System

Android

### AppFlowy Editor Version(s)

2.3.3

### Screenshots

https://github.com/AppFlowy-IO/appflowy-editor/assets/87962825/f6b0023f-be99-4ca5-98c0-3ac1d603fb40

![image](https://github.com/AppFlowy-IO/appflowy-editor/assets/87962825/ffc7db30-327b-4d45-bb6f-8dc47757408f)

## Additional Information
- This issue seems to be specific to the handling of Korean input, particularly when alternating between consonants and vowels.
- [Include any other information that might be helpful for diagnosing the issue, such as screenshots, error logs, or observations regarding when the issue occurs.]

kyungilcho:
new version does not fix this issue at all. 

https://github.com/user-attachments/assets/5cddc7be-7b68-4dc0-ba33-173d252e73a7

galaxy S20

appflowy editor versions: 3.1.0

kyungilcho:
I think that my PR was not a bad choice...

PR #756, opened 2024-04-04
fix: Issue with Korean Input: Loss of TextEditingRange When Alternating Consonant and Vowel

kyungilcho:
#755 

Previously, when Korean is editing, composing is not working correctly. composing region changed by the framework.


LucasXu0:
@kyungilcho On which platform did you test it?

kyungilcho:
It was Android, Galaxy S7 tablet and Galaxy S20 smart phone.

kyungilcho:
I have tried many different platforms(iOS tablet, iOS(17.x)) and I have faced same issue. Now I fixed the issue please test it and give me feed back.

LucasXu0:
The latest version 3.0.0 should include the fix for the text range composition issue.

PR #766, opened 2024-04-11
feat:add table style

q200892907:
add table style
Default width and height can be customized


test/new/block_component/table_block_component/table_node_test.dart
```diff
@@ -224,7 +224,7 @@ void main() {
         ['1', '2'],
         ['3', '4'],
       ]);
-      const config = TableConfig();
+      var config = TableConfig();
```
LucasXu0, line 227: use final instead of var here.

PR #781, opened 2024-04-25
fix:replace_text

q200892907:
fix https://github.com/AppFlowy-IO/AppFlowy/issues/5189

https://github.com/AppFlowy-IO/appflowy-editor/assets/4517301/b0e8b370-6503-4e6a-a474-439c686ac3f7

### DIFF
--- a/lib/src/core/transform/transaction.dart
+++ b/lib/src/core/transform/transaction.dart
@@ -1,5 +1,3 @@
-import 'dart:math';
-
 import 'package:appflowy_editor/appflowy_editor.dart';
 
 /// A [Transaction] has a list of [Operation] objects that will be applied
@@ -17,6 +15,7 @@ class Transaction {
 
   /// The operations to be applied.
   final List<Operation> _operations = [];
+
   List<Operation> get operations {
     if (markNeedsComposing) {
       // compose the delta operations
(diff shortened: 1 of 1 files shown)


q200892907:
> @q200892907 Your change is incorrect. It will cause the input text to lose its format.

I don't understand what you mean. I tested all the test cases and passed them, and there was no problem with input verification.

q200892907:
@LucasXu0 I understand what it means. I will modify the code later and add a blacklist to handle the format.

q200892907:
```
 inlineAttributesBlacklist.addAll([
      InlineMathEquationKeys.formula,
    ]); 
```
Add blacklist to solve text input problem

lib/src/core/transform/transaction.dart
```diff
@@ -346,6 +349,12 @@ extension TextTransaction on Transaction {
     var newAttributes = attributes;
     if (index != 0 && attributes == null) {
       newAttributes = delta.slice(max(index - 1, 0), index).first.attributes;
+      if (newAttributes != null) {
```
LucasXu0, line 352: Using `sliceAttributes` instead of `delta.slice(max(index - 1, 0), index).first.attributes` will be sufficient.

The `sliceAttributes` function will only slice the keys in `AppFlowyRichTextKeys.supportSliced`, which is a whitelist containing the attributes supported for slicing.

PR #785, opened 2024-04-26
fix: numbered list doesn't increased when exporting html

LucasXu0:


LucasXu0:
closes https://github.com/AppFlowy-IO/AppFlowy/issues/5216

PR #788, opened 2024-05-03
fix: update selection from the delta

LucasXu0:
In the common case, the cursor will be placed after the composing text. So there's no need to update the selection manually.

However, on Windows and for some IMEs, the cursor will be placed at the front of the composing text. So updating the selection from the deltas is sufficient.


test/new/editor_component/service/shortcuts/character_shortcuts/back_slash_igore_format_test.dart
Xazin, line 1: Should we just delete this file.

I assume we will have to find a better solution for the markdown ignore anyhow, and the behavior might be quite different.

LucasXu0: Agreed. I have deleted it. It's not a good approach to ignore the markdown syntax here.

PR #794, opened 2024-05-12
math equation

Conceptatlead:
- Added math equation block component to allow users to insert math equations into the editor.
- Included the necessary assets, such as the math equation icon.
- Modified relevant files to integrate the math equation functionality into the editor.
- Made adjustments to text formatting operations to accommodate math equations.
- Updated dependencies in the pubspec.yaml file as needed.


Xazin:
The plugin should be added to the `AppFlowy-Plugins` repository in the `AppFlowy-Editor-Plugins` package, and not directly in the Editor package itself.

Can you move the PR to this [repository](https://github.com/AppFlowy-IO/appflowy-plugins)https://github.com/AppFlowy-IO/appflowy-plugins?

Conceptatlead:
@Xazin  The reason i have integrated it in main because im in need for fully integrated solution, it also has support for importing and exporting math equation in markdown. Currently working on elements like outline, inline math ,call out etc, will push to plugin once done

PR #822, opened 2024-06-18
[fix] Fixed the issue where the handwriting input method was unusable…

ChenTianSaber:
Fixed the issue where the handwriting input method was unusable on iOS and unable to delete on Android.

My users reported a bug in the handwriting input feature, and I've fixed it.

https://github.com/AppFlowy-IO/appflowy-editor/assets/16030568/635e9215-60c6-4535-818e-3f86cf70d6bf

### DIFF
--- a/lib/src/editor/editor_component/service/ime/non_delta_input_service.dart
+++ b/lib/src/editor/editor_component/service/ime/non_delta_input_service.dart
@@ -188,8 +188,10 @@ class NonDeltaTextInputService extends TextInputService with TextInputClient {
           start: composingTextRange!.start,
           end: delta.composing.end,
         );
+        return;
       } else {
         composingTextRange = delta.composing;
+        return;
       }
     }
 
@@ -198,12 +200,11 @@ class NonDeltaTextInputService extends TextInputService with TextInputClient {
             PlatformExtension.isMacOS) &&
         delta is TextEditingDeltaNonTextUpdate) {
       composingTextRange = delta.composing;
+      return;
     }
 
     // solve the issue where the Chinese IME doesn't continue deleting after the input content has been deleted.
-    if (composingTextRange?.isCollapsed ?? false) {
-      composingTextRange = TextRange.empty;
-    }
+    composingTextRange = TextRange.empty;
   }
 }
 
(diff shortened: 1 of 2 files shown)


LucasXu0:
```dart
  void _updateComposing(TextEditingDelta delta) {
    if (delta is TextEditingDeltaNonTextUpdate) {
      composingTextRange = delta.composing;
    } else {
      composingTextRange = composingTextRange != null &&
              composingTextRange!.start != -1 &&
              delta.composing.end != -1
          ? TextRange(
              start: composingTextRange!.start,
              end: delta.composing.end,
            )
          : delta.composing;
    }
  }
```

@ChenTianSaber I did some changes on the `_updateComposing` function. You may can refer to it.

LucasXu0:
@ChenTianSaber By the way, what did you mean by the issue "unable to delete on Android."?

ChenTianSaber:
> @ChenTianSaber By the way, what did you mean by the issue "unable to delete on Android."?

@LucasXu0 
Please watch the following video. The steps to reproduce the issue are as follows:

https://github.com/AppFlowy-IO/appflowy-editor/assets/16030568/4c91dcd1-853a-490f-850e-b230dd833624

1. Use the handwriting input method on Android.
2. After inputting, keep the text in the composing state.
3. At this point, clicking the delete button has no effect.

The reason for this issue is that when the text is in the composing state, enableShortcuts is set to false. As a result, clicking the delete button only triggers the onKeyEvent event, which gets ignored because enableShortcuts is false.

ChenTianSaber:
> ```dart
>   void _updateComposing(TextEditingDelta delta) {
>     if (delta is TextEditingDeltaNonTextUpdate) {
>       composingTextRange = delta.composing;
>     } else {
>       composingTextRange = composingTextRange != null &&
>               composingTextRange!.start != -1 &&
>               delta.composing.end != -1
>           ? TextRange(
>               start: composingTextRange!.start,
>               end: delta.composing.end,
>             )
>           : delta.composing;
>     }
>   }
> ```
> 
> @ChenTianSaber I did some changes on the `_updateComposing` function. You may can refer to it.

Thanks, I will try this code.

LucasXu0:
@ChenTianSaber I also tested the deletion issue on Android. It seems working fine if using the code snippet I posted without changing the keyboard service.

ChenTianSaber:
> @ChenTianSaber I also tested the deletion issue on Android. It seems working fine if using the code snippet I posted without changing the keyboard service.

image

@LucasXu0 I used your code and rolled back the _keyboard_service_widget.dart_ code, but the issue still exists on Android.

Are you using the Android handwriting input method? The problem only occurs with the handwriting input method. As I mentioned above, when enableShortcuts = false and the delete key is pressed, it only triggers _onKeyEvent. At this point, the event is ignored.

LucasXu0:
@ChenTianSaber I'll submit a new pull request to resolve the backspace issue in Android handwriting input.

PR #850, opened 2024-07-27
feat: support single column slash menu

appflowy:


appflowy:
Uploading Screen Recording 2024-07-27 at 14.43.00.mov…

PR #860, opened 2024-08-08
fix: enable converting multilne to list

zoli:
currently, it's only possible to convert a single line to a bullet or numbered list.
changed the active check function to also show list items in toolbar when multiple text lines are selected.

closes https://github.com/AppFlowy-IO/AppFlowy/issues/3181


black-puppydog:
Sorry to necropost here, but did this get removed at some point? I should see this when selecting multiple lines of text, right? This is what I see...

image

PR #863, opened 2024-08-09
feat: paste list children nodes

zoli:
also, export copy paste extension so we can use it in AppFlowy.


zoli:
This should get merged before https://github.com/AppFlowy-IO/AppFlowy/pull/5917.

PR #875, opened 2024-09-04
feat: optimize the table node parse steps and add logs

LucasXu0:
Remove the assertion and forced unwrapping, make the code more production-friendly, and add more logging.

### DIFF
--- a/lib/src/editor/block_component/table_block_component/table_node.dart
+++ b/lib/src/editor/block_component/table_block_component/table_node.dart
@@ -12,44 +12,59 @@ class TableNode {
   TableNode({
     required this.node,
   }) : _config = TableConfig.fromJson(node.attributes) {
-    assert(node.type == TableBlockKeys.type);
-    assert(node.attributes.containsKey(TableBlockKeys.colsLen));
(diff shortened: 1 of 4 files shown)


LucasXu0:
@zoli FYI, you may be interested in this PR. :)

lib/src/editor/block_component/table_block_component/table_node.dart
```diff
@@ -12,44 +12,59 @@ class TableNode {
...
+        colsLen is! int ||
+        rowsLen is! int) {
+      Log.editor.debug(
+        'TableNode: colsLen or rowsLen is not an integer or null',
+      );
+      return;
+    }
 
-    for (var i = 0; i < colsCount; i++) {
+    if (node.children.length != colsLen * rowsLen) {
```
zoli, line 34: Why don't you just call validate on the node (maybe with some customization on the validate function)?

zoli:
> @zoli FYI, you may be interested in this PR. :)

Thanks. I saw an issue regarding the table node the root of that problem hasn't been found yet. right?

PR #876, opened 2024-09-04
feat: support more link types

SillyCoon:
According to https://github.com/AppFlowy-IO/AppFlowy/issues/5146 I added support of `mailto` and `file` URIs.

It's also possible to use built-in [Uri.tryParse](https://api.flutter.dev/flutter/dart-core/Uri/tryParse.html) to support any URI, but I'm not sure if this is a good idea since it allows any schema (e.g. `javascript` too) and probably should be validated more. 

So for now I left it simple.

### DIFF
--- a/lib/src/editor/util/link_util.dart
+++ b/lib/src/editor/util/link_util.dart
@@ -1,11 +1,15 @@
 import 'package:string_validator/string_validator.dart';
 
 bool isUri(String text) {
-  return isURL(text) || isMailTo(text);
+  return isURL(text) || isMailTo(text) || isFile(text);
 }
 
 bool isMailTo(String text) {
   const mailToPrefix = 'mailto:';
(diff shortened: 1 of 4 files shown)


lib/src/editor/util/link_util.dart
```diff
@@ -0,0 +1,15 @@
...
+
+bool isMailTo(String text) {
+  const mailToPrefix = 'mailto:';
+  return text.startsWith(mailToPrefix) &&
+      isEmail(text.substring(mailToPrefix.length));
+}
+
+bool isFile(String text) {
+  return text.startsWith('file:/');
+}
```
LucasXu0, line 15:
```suggestion
bool isUri(String text) {
  final lowerText = text.toLowerCase();
  return isURL(text) || 
         lowerText.startsWith('mailto:') || 
         lowerText.startsWith('file:');
}
```

PR #878, opened 2024-09-08
fix: regain focus after navigation arrows are used

MayurSMahajan:
Solves: 
[#5491](https://github.com/AppFlowy-IO/AppFlowy/issues/5491)
[#5548](https://github.com/AppFlowy-IO/AppFlowy/issues/5548)
[#6211](https://github.com/AppFlowy-IO/AppFlowy/issues/6211)

The core issue is that the **find** `InputField` loses focus after match navigation keys are pressed. 
Let me explain:
The **find** `InputField` has listens for `Enter` key press and navigates to the next found match. 
Thus if the user were to find some text and keep pressing `Enter` then there will be no problem.

But
the main problem occurs when the user presses one of the navigating keys button, which causes the `InputField` to lose focus.

And when the input field loses focus, pressing `Enter` or `Return` causes the editor to delete the selection instead of navigating to the next match.

Thus this PR applies a workaround where the `InputField` gains focus even after navigating keys are pressed.

---

There is the third case where, the user deliberately makes a selection in the document and then pressing `Enter` should ideally modify the document.


MayurSMahajan:
Here's video after some testing:


https://github.com/user-attachments/assets/cbce6b82-3e0d-4038-8a5b-190a63d638f9

Issue #884, opened 2024-09-10
[Bug] Document.blank().first or last throw exception IterableElementError.noElement();


Linij:
### Bug Description

**Related file path:**
lib/src/core/document/document.dart 

**Error code at line(78 - 87):**
 ```
 /// First node of the document.
Node? get first => root.children.first;

/// Last node of the document.
Node? get last {
  var last = root.children.last;
  while (last.children.isNotEmpty) {
    last = last.children.last;
  }
  return last;
}
```

### How to Reproduce

just run code `Document.blank().first` or `Document.blank().last`  while see.

### Expected Behavior

 throw exception IterableElementError.noElement();

### Operating System

macos 

### AppFlowy Editor Version(s)

3.2.0

### Screenshots

<img width="860" alt="image" src="https://github.com/user-attachments/assets/69862057-a863-4541-b35b-81a6ab7a7c39">

PR #885, opened 2024-09-10
fix: document getting first or last node exception

Linij:
fix issue https://github.com/AppFlowy-IO/appflowy-editor/issues/884

### DIFF
--- a/lib/src/core/document/document.dart
+++ b/lib/src/core/document/document.dart
@@ -75,15 +75,15 @@ class Document {
   final Node root;
 
   /// First node of the document.
-  Node? get first => root.children.first;
+  Node? get first => root.children.firstOrNull;
 
   /// Last node of the document.
   Node? get last {
-    var last = root.children.last;
-    while (last.children.isNotEmpty) {
-      last = last.children.last;
+    Node? current = root.children.lastOrNull;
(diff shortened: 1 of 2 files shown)


lib/src/core/document/document.dart
```diff
@@ -75,10 +75,14 @@ class Document {
   final Node root;
 
   /// First node of the document.
-  Node? get first => root.children.first;
+  Node? get first => root.children.isNotEmpty ? root.children.first : null;
```
LucasXu0, line 78:
```suggestion
  Node? get first => root.children.firstOrNull;
```

lib/src/core/document/document.dart
```diff
@@ -75,10 +75,14 @@ class Document {
...
 
   /// Last node of the document.
   Node? get last {
+    if (root.children.isEmpty) {
```
LucasXu0, line 82:
```dart
  Node? get last {
    Node? current = root.children.lastOrNull;
    while (current != null && current.children.isNotEmpty) {
      current = current.children.last;
    }
    return current;
  }
```

LucasXu0:
@Linij Thanks for your contribution. Please update your comment.

Linij:
hi,i accept the suggesstion code, and update the commit.

PR #888, opened 2024-09-12
feat: Allow configuring the auto scroll edge offset

snipd-min:
When using the `MobileFloatingToolbar` with a markdown editor not in full screen design (e.g. if a docked footer is present on the screen), the default auto scroll edge might not work and would lead to the following assertion error:

```
  assert(
      globalRect.size.width >= _dragTargetRelatedToScrollOrigin.size.width &&
          globalRect.size.height >=
              _dragTargetRelatedToScrollOrigin.size.height,
      'Drag target size is larger than scrollable size, which may cause bouncing',
    );
```

In this case it might be useful to allow configuring the edge offset.


LucasXu0:
@snipd-min Thanks for contribution. Please sign the CLA :)

snipd-min:
Done :)

PR #896, opened 2024-09-20
feat: support moving a block to become a child of another block

LucasXu0:
Expose the `builder` parameters in `renderDropTargetForOffset` function to allow developers to customize the drag area widget.

https://github.com/user-attachments/assets/2ee8de7d-ffe3-48c1-9cc3-3248d78b3ed8

### DIFF
--- a/example/lib/pages/drag_to_reorder_editor.dart
+++ b/example/lib/pages/drag_to_reorder_editor.dart
@@ -325,13 +325,8 @@ class _DragToReorderActionState extends State<DragToReorderAction> {
   ) {
     final node = data.targetNode;
     final selectable = node.selectable;
-
-    if (selectable == null) {
-      return null;
-    }
-
-    final renderBox = selectable.context.findRenderObject() as RenderBox?;
-    if (renderBox == null) {
+    final renderBox = selectable?.context.findRenderObject() as RenderBox?;
+    if (selectable == null || renderBox == null) {
       return null;
     }
 
--- a/example/lib/pages/drag_to_reorder_editor.dart
+++ b/example/lib/pages/drag_to_reorder_editor.dart
@@ -3,6 +3,10 @@ import 'package:flutter/material.dart';
 import 'package:google_fonts/google_fonts.dart';
 import 'package:provider/provider.dart';
 
+enum HorizontalPosition { left, center, right }
+
+enum VerticalPosition { top, middle, bottom }
+
 class DragToReorderEditor extends StatefulWidget {
   const DragToReorderEditor({
     super.key,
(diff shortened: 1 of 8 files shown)


example/lib/pages/drag_to_reorder_editor.dart
```diff
@@ -236,4 +251,125 @@ class _DragToReorderActionState extends State<DragToReorderAction> {
...
+    final selectable = node.selectable;
+
+    if (selectable == null) {
+      return const SizedBox.shrink();
+    }
+
+    final renderBox = selectable.context.findRenderObject() as RenderBox?;
+    if (renderBox == null) {
+      return const SizedBox.shrink();
+    }
```
Xazin, line 266:
```suggestion
    final renderBox = selectable?.context.findRenderObject() as RenderBox?;
    if (selectable == null || renderBox == null) {
      return const SizedBox.shrink();
    }
```

example/lib/pages/drag_to_reorder_editor.dart
```diff
@@ -236,4 +251,114 @@ class _DragToReorderActionState extends State<DragToReorderAction> {
...
+    final selectable = node.selectable;
+
+    if (selectable == null) {
+      return null;
+    }
+
+    final renderBox = selectable.context.findRenderObject() as RenderBox?;
+    if (renderBox == null) {
+      return null;
+    }
```
Xazin, line 330:
```suggestion
    final selectable = node.selectable;
    final renderBox = selectable?.context.findRenderObject() as RenderBox?;
    if (selectable == null || renderBox == null) {
      return null;
    }
```

Xazin (approved):

LucasXu0:
```dart
  Future<void> _moveNodeToNewPosition(
    Node node,
    Path? acceptedPath,
    Offset dragOffset,
  ) async {
    if (acceptedPath == null) return;

    final editorState = context.read<EditorState>();
    final targetNode = editorState.getNodeAtPath(acceptedPath);
    if (targetNode == null) return;

    final position = _getPosition(context, targetNode, dragOffset);
    if (position == null) return;

    final (verticalPosition, horizontalPosition, _) = position;
    Path newPath = targetNode.path;

    // Determine the new path based on drop position
    // For VerticalPosition.top, we keep the target node's path
    if (verticalPosition == VerticalPosition.bottom) {
      newPath = horizontalPosition == HorizontalPosition.left
          ? newPath.next // Insert after target node
          : newPath.child(0); // Insert as first child of target node
    }

    // Check if the drop should be ignored
    if (_shouldIgnoreDrop(node, newPath)) {
      debugPrint(
        'Drop ignored: node($node, ${node.path}), path($acceptedPath)',
      );
      return;
    }

    final realNode = widget.blockComponentContext.node;
    debugPrint('Moving node($realNode, ${realNode.path}) to path($newPath)');

    // Perform the node move operation
    final transaction = editorState.transaction;
    transaction.moveNode(newPath, realNode);
    await editorState.apply(transaction);
  }
```

@Xazin I optimized the moving logic to support moving a paragraph as another block's child. However, there are still some corner cases that should be handled, e.g., checking if the node can contain a child.

PR #903, opened 2024-09-26
feat: Multi list indentation at once

Ahad-patel:
Closes : https://github.com/AppFlowy-IO/AppFlowy/issues/6255

This PR solves the issue of mulitple indentation of array lists at once. 

https://github.com/user-attachments/assets/223c1378-8cd4-4eb1-ac1a-62290232d9ad


Ahad-patel:
Hey, @Xazin @LucasXu0  

I have added the feature for mulit list indentation at once, some of the tests are failing can you help me solve these issues?

lib/src/editor/block_component/base_component/indent_command.dart
gouravbambhore, line 1: Good contribution !

Ahad-patel: Thanks @gouravbambhore,

but some of the checks are failing can you help me resolve it?

lib/src/editor/block_component/base_component/indent_command.dart
```diff
@@ -23,41 +23,64 @@ final CommandShortcutEvent indentCommand = CommandShortcutEvent(
 
 bool isIndentable(EditorState editorState) {
   final selection = editorState.selection;
-  if (selection == null || !selection.isSingle) {
```
LucasXu0, line 26: This condition `!selection.isSingle` will prevent the single line from being indented.

LucasXu0:
@Ahad-patel There's one more case that should be considered.

The `getNodesInSelection` function will return all the nodes, including the children. For example, the tab command doesn't work in the case shown in the attached screenshot. When judging if the nodes are indentable, we just need to consider the first level and filter out their children nodes.

And please add more tests to cover this case.

Screenshot 2024-09-30 at 10 23 08

Ahad-patel:
Hey @LucasXu0, 

I have updated outdent_command file according to your changes, I have also added tests cases for indent_command and outdent_command.

 I noticed that the outdent_command test file was named "outdent_handler_test.dart" ,since other test files were named according to their file name, so I renamed it to "outdent_command_test.dart".   

I also removed isAllOnSameLevel because we are ignoring child nodes, so the isAllOnSameLevel method will always be true.

Are there any more remaining tasks in this PR?

LucasXu0:
> I noticed that the outdent_command test file was named "outdent_handler_test.dart" ,since other test files were named according to their file name, so I renamed it to "outdent_command_test.dart".

Good catch.

> Are there any more remaining tasks in this PR?

I'm checking the newly added code and will comment on the PR.

test/new/service/shortcuts/command_shortcut_events/outdent_command_test.dart
```diff
@@ -212,4 +212,167 @@ void main() async {
...
     );
   });
+
+  group('outdentCommand (multi-line) - widget test multi-line outdent padding',
+      () {
+    testWidgets("indent two items of same level then outdent it",
+        (tester) async {
+      const text = 'Welcome to Appflowy 😁';
+      final editor = tester.editor
+        // final editor = tester.editor..addNode(paraNode);
```
LucasXu0, line 222: remove this line.

test/new/service/shortcuts/command_shortcut_events/outdent_command_test.dart
```diff
@@ -212,4 +212,167 @@ void main() async {
...
+      expect(editor.nodeAtPath([0, 0]), null);
+
+      await editor.dispose();
+    });
+
+    testWidgets("ignore if any bullet list of level 1 selected",
+        (tester) async {
+      const text = 'Welcome to Appflowy 😁';
+      final editor = tester.editor
+        // final editor = tester.editor..addNode(paraNode);
```
LucasXu0, line 304: remove this line

test/new/service/shortcuts/command_shortcut_events/outdent_command_test.dart
Ahad-patel: done

test/new/service/shortcuts/command_shortcut_events/outdent_command_test.dart
Ahad-patel: done

LucasXu0:
> There is a selection before and after the line when this is in selection editorState.getNodesInSelection will give both the nodes.

It works as intended. You can see the toolbar location. Actually, the previous line has been selected as well. An improvement we can make here is to add a visual selection area for the previous line.

Screenshot 2024-10-04 at 11 10 12

Ahad-patel:
> An improvement we can make here is to add a visual selection area for the previous line.

Should I make an new Issue and new PR for this ?

Xazin:
> > There is a selection before and after the line when this is in selection editorState.getNodesInSelection will give both the nodes.
> 
> It works as intended. You can see the toolbar location. Actually, the previous line has been selected as well. An improvement we can make here is to add a visual selection area for the previous line.
> 


Normally if you have the cursor before the last character, and you use `Shift + arrow right`, it'll select the last character without the newline (so not the next line). This should be seen as a bug.

Ahad-patel:
> Normally if you have the cursor before the last character, and you use `Shift + arrow right`, it'll select the last character without the newline (so not the next line). This should be seen as a bug.

I have created a new issue for this https://github.com/AppFlowy-IO/AppFlowy/issues/6500

PR #915, opened 2024-10-08
fix: do not highlight inactive items on empty selection

Xazin:
Closes: [#1681](https://github.com/AppFlowy-IO/AppFlowy/issues/1681)

This is simply an improvement to the behavior, there is nothing to "apply" formatting options on, so should we even show these options? We can improve that later..

## Before:

https://github.com/user-attachments/assets/d38511a3-fe40-4979-8f69-b34a565f5c03

## After:

https://github.com/user-attachments/assets/622505e3-40dc-41b4-afe1-b8fe3d0c3b65

### DIFF
--- a/lib/src/render/toolbar/toolbar_item.dart
+++ b/lib/src/render/toolbar/toolbar_item.dart
@@ -1,5 +1,6 @@
-import 'package:appflowy_editor/appflowy_editor.dart';
 import 'package:flutter/material.dart' hide Overlay, OverlayEntry;
+
+import 'package:appflowy_editor/appflowy_editor.dart';
 
 typedef ToolbarItemEventHandler = void Function(
   EditorState editorState,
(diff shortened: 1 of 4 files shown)


lib/src/render/toolbar/toolbar_item.dart
```diff
@@ -1,6 +1,7 @@
-import 'package:appflowy_editor/appflowy_editor.dart';
 import 'package:flutter/material.dart' hide Overlay, OverlayEntry;
 
+import 'package:appflowy_editor/appflowy_editor.dart';
```
LucasXu0, line 3: Looks like you use a different formatter than I do. Every time I format the code, these imports will be reverted.

Xazin: I use this extension https://marketplace.visualstudio.com/items?itemName=luanpotter.dart-import for cleaner imports, makes them easier to read in my opinion.

I don't think it matters much that we keep changing them, what I don't understand is why the built-in formatter is so fond of changing it back 😂.

PR #917, opened 2024-10-08
fix: ctrl+delete does not delete newlines

Xazin:
Closes: [#3989](https://github.com/AppFlowy-IO/AppFlowy/issues/3989)


lib/src/editor/editor_component/service/shortcuts/command/remove_word_command.dart
```diff
@@ -96,8 +97,12 @@ CommandShortcutEventHandler _deleteRightWordCommandHandler = (editorState) {
     return KeyEventResult.ignored;
   }
 
+  if (selection.start.offset == delta.length) {
```
LucasXu0, line 100: Should the selection be collapsed?

Issue #927, opened 2024-10-14
[Bug] _deleteBackward deleteRange start must > 0 


Linij:
### Bug Description

<img width="1598" alt="image" src="https://github.com/user-attachments/assets/77c686fc-db3e-4d76-ac21-d825b8a12d3d">
<img width="1059" alt="image" src="https://github.com/user-attachments/assets/c57eeeea-388e-4b9e-9021-13a7f7b334cb">

### How to Reproduce

Continue clicking BackSPACE when there is no content on the first line.

### Expected Behavior

Judgment of modification and deletion scope conditions

### Operating System

Macos

### AppFlowy Editor Version(s)

4.0.0

### Screenshots

<img width="2584" alt="image" src="https://github.com/user-attachments/assets/e1ef52c1-5e66-4867-b0c1-5874144ede5c">

PR #928, opened 2024-10-14
fix: _deleteBackward deleteRange start must > 0

Linij:
issue: #927

### DIFF
--- a/lib/src/editor/editor_component/service/ime/non_delta_input_service.dart
+++ b/lib/src/editor/editor_component/service/ime/non_delta_input_service.dart
@@ -207,7 +207,7 @@ class NonDeltaTextInputService extends TextInputService with TextInputClient {
       final selection = currentTextEditingValue.selection;
       final deleteRange = selection.isCollapsed
           ? TextRange(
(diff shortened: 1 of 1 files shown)


lib/src/editor/editor_component/service/ime/non_delta_input_service.dart
```diff
@@ -207,7 +207,7 @@ class NonDeltaTextInputService extends TextInputService with TextInputClient {
       final selection = currentTextEditingValue.selection;
       final deleteRange = selection.isCollapsed
           ? TextRange(
-              start: selection.start - 1,
+              start: max(0, selection.start - 1),
```
LucasXu0, line 210: Should check the selection status instead of using `max` here. If the selection is invalid, skip the selector.

saif-ellafi:
Hello @Linij @LucasXu0 - thank you for this fix. Tested on my system and it works fine.
However, deleting a Table, when the Table is the last thing being removed, also leaves the Editor in the same (broken) state.
Could this fix be extended to those other areas perhaps?

Issue #948, opened 2024-10-28
[Bug] Deleting the last empty entry crashes the Editor


saif-ellafi:
### Bug Description

In my AppFlowyEditor, if I have an empty entry, and I press delete (i.e. delete the empty line), then the Editor crashes.
The Editor also crashes if the editor is fully blank. In MacOS, I am allowed to delete an empty line by pressing backspace. This also happens when I delete a Table and the table was the last thing I had in the Editor. 

The MacOS version issue may be the one fixed here: https://github.com/AppFlowy-IO/appflowy-editor/issues/927 - not sure about the deleting table version of this issue

```
════════ Exception caught by gesture ═══════════════════════════════════════════
Invalid argument(s): 0
════════════════════════════════════════════════════════════════════════════════
Restarted application in 1,761ms.
flutter: unhandled element <style/>; Picture key: Svg loader
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: RangeError (start): Invalid value: Only valid value is 0: -1
#0      RangeError.checkValidRange (dart:core/errors.dart:360:7)
errors.dart:360
#1      _StringBase.substring (dart:core-patch/string_patch.dart:418:27)
string_patch.dart:418
#2      TextEditingDeltaDeletion.textDeleted (package:flutter/src/services/text_editing_delta.dart:314:37)
text_editing_delta.dart:314
#3      TextEditingDeltaDeletion.debugFillProperties (package:flutter/src/services/text_editing_delta.dart:333:63)
text_editing_delta.dart:333
#4      DiagnosticableNode.builder.<anonymous closure> (package:flutter/src/foundation/diagnostics.dart:2837:17)
diagnostics.dart:2837
#5      DiagnosticableNode.builder (package:flutter/src/foundation/diagnostics.dart:2840:8)
diagnostics.dart:2840
#6      DiagnosticableNode.getProperties (package:flutter/src/foundation/diagnostics.dart:2854:105)
diagnostics.dart:2854
```

Even worse, if the last thing I delete is a Table, then the Editor crashes, but also the app does not resume, and needs a full app restart to use it. This is the error when the last entry fully deleted is a table:

```
════════ Exception caught by gesture ═══════════════════════════════════════════
The following ArgumentError was thrown while handling a gesture:
Invalid argument(s): 0

When the exception was thrown, this was the stack:
#0      int.clamp (dart:core-patch/integers.dart:280:7)
integers.dart:280
[...]

PR #951, opened 2024-10-29
fix: deleting the last table being the last node of the document, places an empty node to avoid the editor to crash

saif-ellafi:
Dear all,

First ever pull request, probably not great, but it solves https://github.com/AppFlowy-IO/appflowy-editor/issues/948 , by preventing an empty state Document that bricks the editor after deleting a table, when such table was the last thing on the Document, leaving the editor in an broken state.

In my app, since users bricked their save games with this bug, I am detecting (after loading from file), whether there is any Document on empty state, and if so, I overwrite them with a blank Document. This is to recover into a healthy state, with this patch preventing it from happening again (from this, or any other cause that leads to the same state).

Any suggestions welcome, it just gets the work done for now.

Best regards
Saif

### DIFF
--- a/lib/src/editor/block_component/table_block_component/table_action.dart
+++ b/lib/src/editor/block_component/table_block_component/table_action.dart
@@ -214,6 +214,10 @@ void _deleteCol(Node tableNode, int col, EditorState editorState) {
       colsLen = tableNode.attributes[TableBlockKeys.colsLen];
 
   if (colsLen == 1) {
+    if (editorState.document.root.children.length == 1) {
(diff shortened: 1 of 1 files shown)


LucasXu0:
Hi, @saif-ellafi. Please sign the CLA.

saif-ellafi:
@LucasXu0 - done, I noticed a lint check did not pass. Need to investigate.

saif-ellafi:
I think the lint error is regarding the commit message, no? any advice is welcome.

PR #968, opened 2024-11-15
chore: don't attach text input service when editor isn't editable

richardshiue:


richardshiue:
Closing as `disableKeyboardService` can be used.

PR #975, opened 2024-11-28
fix: the cursor cannot exit the table using up and down arrow keys

asjqkkkk:
Before: 

https://github.com/user-attachments/assets/4e63c6e1-be82-496e-8d5c-0ca5f956f908

Aftre:

https://github.com/user-attachments/assets/c63e379a-cd83-40e9-847c-db4095448f0e

At present, it has only resolved the issue of the cursor jumping out with the up and down arrow keys; there are other issues, such as: using arrow keys to move from other nodes into a table causing the cursor to disappear. These may need to be addressed through other pull requests.

### DIFF
--- a/lib/src/editor/editor_component/service/shortcuts/character/slash_command.dart
+++ b/lib/src/editor/editor_component/service/shortcuts/character/slash_command.dart
@@ -5,6 +5,15 @@ import 'package:appflowy_editor/src/editor/util/platform_extension.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 
+const Set<String> _defaultSupportSlashMenuNodeTypes = {
+  ParagraphBlockKeys.type,
+  HeadingBlockKeys.type,
+  TodoListBlockKeys.type,
+  BulletedListBlockKeys.type,
+  NumberedListBlockKeys.type,
+  QuoteBlockKeys.type,
+};
+
 /// Show the slash menu
 ///
 /// - support
@@ -26,6 +35,7 @@ CharacterShortcutEvent customSlashCommand(
   bool deleteKeywordsByDefault = false,
   bool singleColumn = true,
   SelectionMenuStyle style = SelectionMenuStyle.light,
+  Set<String> supportSlashMenuNodeTypes = _defaultSupportSlashMenuNodeTypes,
 }) {
   return CharacterShortcutEvent(
     key: 'show the slash menu',
@@ -37,18 +47,10 @@ CharacterShortcutEvent customSlashCommand(
       deleteKeywordsByDefault: deleteKeywordsByDefault,
       singleColumn: singleColumn,
       style: style,
+      supportSlashMenuNodeTypes: supportSlashMenuNodeTypes,
     ),
   );
 }
-
-final Set<String> supportSlashMenuNodeWhiteList = {
-  ParagraphBlockKeys.type,
-  HeadingBlockKeys.type,
-  TodoListBlockKeys.type,
-  BulletedListBlockKeys.type,
-  NumberedListBlockKeys.type,
-  QuoteBlockKeys.type,
-};
 
 SelectionMenuService? _selectionMenuService;
 Future<bool> _showSlashMenu(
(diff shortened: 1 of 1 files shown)


lib/src/editor/block_component/table_block_component/table_commands.dart
```diff
@@ -146,14 +148,25 @@ CommandShortcutEventHandler _upInTableCellHandler = (editorState) {
...
   final selection = editorState.selection;
   if (_hasSelectionAndTableCell(inTableNodes, selection)) {
-    final nextNode = _getNextNode(inTableNodes, 0, -1);
-    if (_nodeHasTextChild(nextNode)) {
-      final target = nextNode!.childAtIndexOrNull(0)!;
-      final off = target.delta!.length > selection!.start.offset
-          ? selection.start.offset
-          : target.delta!.length;
+    try {
+      final nextNode = _getNextNode(inTableNodes, 0, -1);
```
LucasXu0, line 152: Please use `_getPreviousNode` instead of `_getNextNode`. Also, it would be better to refactor the function `_getPreviousNode` to add named parameters.


```dart
Node? _getPreviousNode(
  Iterable<Node> nodes, {
  int colDiff = 0,
  int rowDiff = 0,
}) {}
```

```dart
final previousNodeInSameColumn = _getPreviousNode(
      inTableNodes,
      rowDiff: 1,
    );
```

```dart
CommandShortcutEventHandler _upInTableCellHandler = (editorState) {
  final inTableNodes = _inTableNodes(editorState);
  if (selection == null ||
      !_hasSelectionAndTableCell(inTableNodes, selection)) {
    return KeyEventResult.ignored;
  }

  final previousNodeInSameColumn = _getPreviousNode(
    inTableNodes,
    rowDiff: 1,
  );
  if (previousNodeInSameColumn != null &&
      _nodeHasTextChild(previousNodeInSameColumn)) {
    final target = previousNodeInSameColumn.childAtIndexOrNull(0);
    final delta = target?.delta;
    if (target == null || delta == null) {
      return KeyEventResult.ignored;
    }
    final offset = delta.length.clamp(
      0,
      selection.start.offset,
    );
    editorState.selectionService.updateSelection(
      Selection.single(
        path: target.path,
        startOffset: offset,
      ),
    );
  }

  return KeyEventResult.handled;
};
```

lib/src/editor/block_component/table_block_component/table_commands.dart
```diff
@@ -165,16 +178,28 @@ CommandShortcutEventHandler _downInTableCellHandler = (editorState) {
   final inTableNodes = _inTableNodes(editorState);
   final selection = editorState.selection;
   if (_hasSelectionAndTableCell(inTableNodes, selection)) {
-    final nextNode = _getNextNode(inTableNodes, 0, 1);
-    if (_nodeHasTextChild(nextNode)) {
-      final target = nextNode!.childAtIndexOrNull(0)!;
-      final off = target.delta!.length > selection!.start.offset
-          ? selection.start.offset
-          : target.delta!.length;
+    try {
```
LucasXu0, line 181: Please help remove force-unwrap operators (!) as well.

LucasXu0:
Hi, @asjqkkkk. Please check the CI. You need to format the code.

asjqkkkk:
Hi @LucasXu0 , I've changed `_getNextNode` to `_getPreviousNode` in `_upInTableCellHandler`, and refactor the function _getPreviousNode to add named parameters

The direction operations of the table have four: up, down, left, and right, and basically, each move will only shift by one cell. The parameters [colDiff+rowDiff] are like x and y coordinates that can represent the four directions, so a single method might be sufficient to obtain the next Node. In this case, the [getPreviousNode] method seems a bit redundant. Additionally, if we were to use [getNextNode and getPreviousNode], there should theoretically be some input restrictions on the parameters [colDiff+rowDiff], such as they cannot be negative numbers. However, if we add these restrictions, it seems unnecessary to use the two int values of [colDiff+rowDiff]. Although [colDiff+rowDiff] can cleverly express directions, it increases the difficulty of reading the code, and it does not impose any input restrictions. In comparison, I might prefer to use an enumeration method to express directions.

LucasXu0:
@asjqkkkk Agreed. Using an enum can improve readability. I have another suggestion: using getCellAtPosition to get the cell from the table. I was refactoring the table code and using these 4 functions to get the previous/next focusable node.

https://github.com/AppFlowy-IO/AppFlowy/pull/6831/files#diff-d3499f42f1d6695407a5ac49d5f3d47ef66979a9db5a03540f80c55eeac0ada4R352

PR #979, opened 2024-12-02
fix: out of text range while composing error

kyungilcho:
Problem:

https://github.com/user-attachments/assets/73f25a31-b82a-495b-9578-d8f638bd3ae7

Re-positioning the cursor in the middle of sentence while composing occurs following error. 

스크린샷 2024-12-02 오후 11 00 51

so I add logic for checking validity of cursor position.


LucasXu0:
@kyungilcho Your PR may fix the composing error, but I think the attached position is still incorrect. The position is empty if the assertion happens.

PR #986, opened 2024-12-10
fix: issue #6808 of appflowy

asjqkkkk:
Fix issue [#6808](https://github.com/AppFlowy-IO/AppFlowy/issues/6808)

Before the fix : 

https://github.com/user-attachments/assets/4a5ea0e4-d7d9-4032-9324-79c204776f71

After the fix:

https://github.com/user-attachments/assets/9f87cba6-c37b-47a0-8019-09a6f4dce27b

### DIFF
--- a/lib/src/editor/editor_component/service/shortcuts/command/paste_command.dart
+++ b/lib/src/editor/editor_component/service/shortcuts/command/paste_command.dart
@@ -86,10 +86,14 @@ extension on EditorState {
   Future<bool> pasteHtml(String html) async {
     final nodes = htmlToDocument(html).root.children.toList();
     // remove the front and back empty line
-    while (nodes.isNotEmpty && nodes.first.delta?.isEmpty == true) {
+    while (nodes.isNotEmpty &&
+        nodes.first.delta?.isEmpty == true &&
+        nodes.first.children.isEmpty) {
       nodes.removeAt(0);
     }
-    while (nodes.isNotEmpty && nodes.last.delta?.isEmpty == true) {
+    while (nodes.isNotEmpty &&
+        nodes.last.delta?.isEmpty == true &&
+        nodes.last.children.isEmpty) {
       nodes.removeLast();
     }
     if (nodes.isEmpty) {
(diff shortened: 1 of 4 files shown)


lib/src/editor/editor_component/service/shortcuts/command/paste_command.dart
```diff
@@ -89,7 +89,9 @@ extension on EditorState {
     while (nodes.isNotEmpty && nodes.first.delta?.isEmpty == true) {
```
LucasXu0, line 89: It may have the same issue at line 89 and needs to check the node.children.

lib/src/plugins/html/html_document_decoder.dart
```diff
@@ -18,7 +18,13 @@ class DocumentHTMLDecoder extends Converter<String, Document> {
...
       return Document.blank(withInitialText: false);
     }
-    final nodes = _parseElement(body.nodes);
+    late Iterable<Node> nodes;
+    if (body.children.length == 1 &&
+        HTMLTags.formattingElements.contains(body.children.first.localName)) {
+      nodes = _parseElement(body.firstChild!.children);
+    } else {
+      nodes = _parseElement(body.nodes);
+    }
```
LucasXu0, line 27:
```suggestion
    final nodes = body.children.length == 1 &&
            HTMLTags.formattingElements.contains(body.children.first.localName)
        ? _parseElement(body.firstChild!.children)
        : _parseElement(body.nodes);
```

test/plugins/html/decoder/_samples.dart
```diff
@@ -557,3 +557,59 @@ const documentSample3 = {
...
   },
 };
+
+const htmlSample4 = '''<meta charset='utf-8'>
+<h1>Problem</h1>
+<ol>
+    <li>Checkout aaaa<span></span><a
+            href="https://docs.appflowy.io/docs/documentation/appflowy/from-source/environment-setup/building-on-macos">啊啊啊啊</a><span></span>asdasdasd</li>
+    <li>222222</li>
+</ol>''';
```
LucasXu0, line 567: It would be better to replace the content with meaningful sentences or words.

PR #1004, opened 2024-12-26
fix: error format of TextEditingDeltaInsertion

asjqkkkk:
To fix https://github.com/AppFlowy-IO/AppFlowy/issues/7049

Before the fix: 

https://github.com/user-attachments/assets/20455a8e-2303-4394-9e52-70d696a01bc4

After the fix:

https://github.com/user-attachments/assets/a89a7e7f-c250-4729-9aba-091ea50f2811

### DIFF
--- a/lib/src/editor/editor_component/service/ime/non_delta_input_service.dart
+++ b/lib/src/editor/editor_component/service/ime/non_delta_input_service.dart
@@ -316,14 +316,18 @@ extension on TextEditingDelta {
   }
 }
 
-extension on TextEditingDeltaInsertion {
-  TextEditingDeltaInsertion format() => TextEditingDeltaInsertion(
-        oldText: oldText << _len,
-        textInserted: textInserted,
-        insertionOffset: insertionOffset - _len,
-        selection: selection << _len,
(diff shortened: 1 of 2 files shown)


lib/src/editor/editor_component/service/ime/non_delta_input_service.dart
```diff
@@ -316,13 +316,15 @@ extension on TextEditingDelta {
...
   TextEditingDeltaInsertion format() => TextEditingDeltaInsertion(
         oldText: oldText << _len,
         textInserted: textInserted,
         insertionOffset: insertionOffset - _len,
-        selection: selection << _len,
-        composing: composing << _len,
+        selection:
+            oldText.startsWith(_whitespace) ? selection << _len : selection,
+        composing:
+            oldText.startsWith(_whitespace) ? composing << _len : composing,
```
LucasXu0, line 327: Please add a test:

1. insert a space into an empty line.
2. delete the space.
3. verify that the result is correct.

asjqkkkk: Already added

PR #1007, opened 2024-12-30
feat: enable debug paint size for mobile drag handle

LucasXu0:
Set `editorState.debugInfo.debugPaintSizeEnabled` to true to render the paint size for the mobile drag handle.

Screenshot 2024-12-30 at 10 27 03


LucasXu0:
closes https://github.com/AppFlowy-IO/appflowy-editor/pull/1007

PR #1037, opened 2025-02-18
fix: editor will crash if deleting all nodes

LucasXu0:
check the nodes length before updating the selection.


yeras-is:
Cmd+A and Delete still crashed editor, can we reopen bug or set a placeholder when content is null ?

PR #1063, opened 2025-03-13
fix: recreate FloatingToolbarWidget after Brightness was changed

asjqkkkk:


lib/src/editor/toolbar/desktop/floating_toolbar_widget.dart
```diff
@@ -107,7 +109,14 @@ class _FloatingToolbarWidgetState extends State<FloatingToolbarWidget> {
     // insert the divider.
     return activeItems
         .splitBetween((first, second) => first.group != second.group)
-        .expand((element) => [...element, placeholderItem])
+        .expand(
+          (element) => [
+            ...element,
+            buildPlaceHolderItemWithColor(
```
LucasXu0, line 115: We can expose a placeholder item builder in the init function so that other developers can customize the placeholder beyond just the color.

PR #1069, opened 2025-03-16
Fixed bug, fixed bug: type 'int' is not a subtype of type 'double?' 

yeras-is:
To reproduce simply convert Document to JSON and Tables will get int types of size instead of double, causing error when reading on Desktop platforms. Somehow Web works fine [Magic].


yeras-is:
mostly I think Web interpretation of Dart doesn't correctly compare **double** and **int** types, that's why it works fine on Web, yet I haven't checked on WASM compilation

PR #1072, opened 2025-03-19
fix: parse table width/height error if the value is int

LucasXu0:
Add dynamic value to double conversion to ensure the width and height values from the table node are doubles.


yeras-is:
still got the same problem after this PR !

Screenshot 2025-03-24 at 15 05 16

PR #1076, opened 2025-03-24
fix: prevent lock after backspacing on select all

saif-ellafi:
Dear all,

I have noticed that in Desktop versions, if you select all (CTRL+A) and press backspace, the editor enters in a locked state. It makes it further unclickable since the document loses all node children.

This PR fixes this.

Best regards,
Saif


LucasXu0:
I've introduced DocumentRule to handle it, but I agree that we should handle it in the backspace command to avoid an uneditable document.

PR #1088, opened 2025-04-02
feat: add RegExp to CharacterShortcutEvent

asjqkkkk:


lib/src/editor/editor_component/service/ime/character_shortcut_event_helper.dart
```diff
@@ -15,7 +15,19 @@ Future<bool> executeCharacterShortcutEvent(
   }
 
   for (final shortcutEvent in characterShortcutEvents) {
-    if (shortcutEvent.character == character &&
+    bool hasMatchRegExp = false;
+    if (shortcutEvent.regExp != null && character != null) {
+      hasMatchRegExp = shortcutEvent.regExp!.hasMatch(character);
```
LucasXu0, line 20: avoid using the `!` syntax.

```dart
final regExp = shortcutEvent.regExp;
if (regExp != null) { 
   // ....
}
```

lib/src/editor/editor_component/service/ime/character_shortcut_event_helper.dart
```diff
@@ -15,7 +15,19 @@ Future<bool> executeCharacterShortcutEvent(
...
+    }
+    if (hasMatchRegExp &&
+        shortcutEvent.handlerWithCharacter != null &&
+        await shortcutEvent.executeWithCharacter(editorState, character!)) {
```
LucasXu0, line 24: also here. `character!`

lib/src/editor/editor_component/service/ime/character_shortcut_event_helper.dart
asjqkkkk: fixed

lib/src/editor/editor_component/service/ime/character_shortcut_event_helper.dart
asjqkkkk: fixed

PR #1184, opened 2026-02-23
fix: prevent documentToMarkdown from mutating source document

LucasXu0:


mic-fomin:
Thanks for landing this! Any plans for a release that includes it? It's affecting downstream apps using documentToMarkdown for auto-save.

PR #1195, opened 2026-04-04
fix: flatten tables so they render in PDF

saif-ellafi:
Dear Appflowy Editor owners,

I was completely failing to export tables in the rich editor into PDF. Getting "TooManyPagesException" issue due to them not fitting in the a4 pages of the export.

You may not like it, but the only solution I found was to plain text the tables to leverage pw.Wrap in a pw.MultiPage.

I hope that even if you don't merge this fix, perhaps you find a better solution one day.

Best regards
Saif


saif-ellafi:
My bad, the issue was related to a specific table being internally corrupted, will hunt the correct bug.

PR #1203, opened 2026-05-10
fix: preserve node identity during type updates

appflowy:
## Summary

Small editor-only change to preserve node identity during safe block type conversions.

Scope is intentionally limited to `appflowy-editor`:
- Add `UpdateNodeTypeOperation` for same-node type/data updates.
- Add `Transaction.updateNodeType` and `Document.updateNodeType`.
- Allow `Node.copyWith` to preserve ids only when explicitly requested.
- Use in-place type updates for markdown shortcut conversions when descendant shape/content is unchanged.
- Emit and apply same-id type diffs from `diffDocuments` / `EditorState`.
- Resolve same-id type updates by stable node id when the operation path is stale.

## Related GitHub Issues

- AppFlowy-IO/AppFlowy#8704 - document fails to open after toggle lists and images; reports `Failed to apply update: block parent ... Type: 8`.
- AppFlowy-IO/AppFlowy#8649 - same `Type: 8` document-open failure on Linux.
- AppFlowy-IO/AppFlowy#8427 - same failure around the 0.11.0 migration boundary.
- AppFlowy-IO/AppFlowy#8124 - earlier multi-platform document corruption report repaired manually, without a code fix.
- AppFlowy-IO/AppFlowy#8414 - adjacent grid/database update error included in the investigation notes.

## Why

The old markdown conversion path replaced a block with insert+delete even when only the block type changed. That regenerated node ids and made downstream clients treat a type conversion as a structural replacement. Desktop/web integration can use this operation to preserve block identity instead.

The user-reported historical context from AppFlowy PR #6516 was also checked before adding coverage. That earlier fix showed two important constraints:
- Conversions into childless block types, such as heading, must flatten nested children instead of keeping them inside the converted block.
- Undo/redo must be verified for both same-id conversions and flattening conversions.

## Test Coverage

Core operation/document coverage:
- `UpdateNodeTypeOperation` serialization, inversion, copy, equality, and hash consistency.
- `Document.updateNodeType` / `Transaction.updateNodeType` preserve node id, children, external values, and metadata.
- `diffDocuments` emits `UpdateNodeTypeOperation` for same-id type changes.
- Remote/local apply handles `UpdateNodeTypeOperation`.
- Stale-path type updates resolve by `nodeId` instead of mutating the wrong path.
- Missing `nodeId` targets are skipped instead of applying to a stale path.
- Undo/redo for `UpdateNodeTypeOperation` preserves parent and child ids.

Markdown conversion coverage:
- Safe no-child conversion: `paragraph -> heading` preserves node id through undo/redo.
- Safe same-shape nested conversion: nested `todo_list -> numbered_list` preserves parent and child ids through undo/redo.
- Childless target conversion: nested `numbered_list -> heading` flattens children into siblings.
- Flattening undo/redo: undo restores the original nested numbered list, redo reapplies the heading plus flattened sibling list items.
- Existing nested `bulleted_list -> heading` flattening behavior remains covered.

## Diff Size

- 13 files changed
- 606 insertions
- 13 deletions
- No desktop, Rust, web, vendored dependency, generated, or lockfile changes

## Verification

- `dart format --output=none --set-exit-if-changed ...`
- `git diff --check`
- `flutter test test/core/transform/operation_test.dart test/core/transform/transaction_test.dart test/new/block_component/heading_block_component/heading_character_shortcut_test.dart test/new/block_component/numbered_list_block_component/numbered_list_character_shortcut_test.dart`

### DIFF
--- a/lib/src/core/document/diff.dart
+++ b/lib/src/core/document/diff.dart
@@ -10,7 +10,18 @@ List<Operation> diffDocuments(Document oldDocument, Document newDocument) {
 List<Operation> diffNodes(Node oldNode, Node newNode) {
   final List<Operation> operations = [];
 
-  if (!_equality.equals(oldNode.attributes, newNode.attributes)) {
+  if (oldNode.type != newNode.type && oldNode.path.isNotEmpty) {
+    operations.add(
+      UpdateNodeTypeOperation(
+        oldNode.path,
+        oldNode.id,
+        newNode.type,
+        oldNode.type,
+        newNode.attributes,
+        oldNode.attributes,
+      ),
+    );
+  } else if (!_equality.equals(oldNode.attributes, newNode.attributes)) {
     operations.add(
       UpdateOperation(oldNode.path, newNode.attributes, oldNode.attributes),
     );
--- a/lib/src/core/document/document.dart
+++ b/lib/src/core/document/document.dart
@@ -157,6 +157,34 @@ class Document {
     return true;
   }
 
+  /// Updates the [Node] type at the given [Path] while preserving its id,
+  /// children, external values, and temporary metadata.
+  bool updateNodeType(Path path, String type, Attributes attributes) {
+    if (path.isEmpty) {
+      return false;
+    }
+
+    final target = nodeAtPath(path);
+    final parent = target?.parent;
+    if (target == null || parent == null) {
+      return false;
+    }
+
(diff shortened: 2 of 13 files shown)


appflowy:
Web https://github.com/AppFlowy-IO/AppFlowy-Web/pull/325

Issue #1204, opened 2026-05-18
[Bug] Flutter 3.44 not supported


sebasrod:
### Bug Description

Error when update to flutter 3.44

../../../AppData/Local/Pub/Cache/hosted/pub.dev/appflowy_editor-6.2.0/lib/src/editor/editor_component/service/ime/delta_input_service.dart:7:7: Error: The non-abstract class 'DeltaTextInputService' is missing implementations for these members:
 - TextInputClient.onFocusReceived
Try to either
 - provide an implementation,
 - inherit an implementation from a superclass or mixin,
 - mark the class as abstract, or
 - provide a 'noSuchMethod' implementation.
class DeltaTextInputService extends TextInputService with DeltaTextInputClient {
      ^^^^^^^^^^^^^^^^^^^^^
/C:/dev/flutter/flutter/packages/flutter/lib/src/services/text_input.dart:1401:8: Context: 'TextInputClient.onFocusReceived' is defined here.
  bool onFocusReceived() => false;
       ^^^^^^^^^^^^^^^
Failed to compile application.

### How to Reproduce

update to flutter 3.44

### Expected Behavior

no error

### Operating System

Windows 11

### AppFlowy Editor Version(s)

6.2

tranthequang:
Thanks for reporting this issue.

I may be able to prepare a focused PR for this, but I need a bit more implementation context first.

Could you please clarify:

1. Which file/module/page is affected?
2. Current behavior
3. Expected behavior
4. Reproduction steps/logs/screenshots if available

Once clarified, I can continue with a targeted fix.

lidissonhugo:
Esta acontecendo este mesmo problema no linux, ao atualizar o flutter para a versão " Flutter (Channel stable, 3.44.1, on Ubuntu 20.04.6 LTS 6.6.114.1-microsoft-standard-WSL2, locale C.UTF-8)" tentar rodar ou fazer um build pela Snap acontece o erro abaixo, não chega nem a completar a compilação:

flutter run

Launching lib/main.dart on Linux in debug mode...
ERROR: ../.pub-cache/hosted/pub.dev/appflowy_editor-6.2.0/lib/src/editor/editor_component/service/ime/delta_input_service.dart:7:7: Error: The non-abstract class 'DeltaTextInputService' is missing implementations for these members:
ERROR:  - TextInputClient.onFocusReceived
ERROR: Try to either
ERROR:  - provide an implementation,
[...]

PR #1205, opened 2026-05-20
Add align support on QuillDeltaEncoder convert, and Fix for Flutter 3.44

rodolfogoulart:
- Add align support when convert from quill to appflowy document.
  - Add handling for 'align' attributes and ensure alignment is applied to nodes. 
- Improve multi-line op.text processing by committing intermediate paragraphs, preserving styles and alignment, and starting a new open node for the last segment. 

- Fix for Flutter 3.44 #1204


salihagic:
Thanks for tackling this @rodolfogoulart. One catch: `NonDeltaTextInputService` (`lib/src/editor/editor_component/service/ime/non_delta_input_service.dart:12`) also mixes in `TextInputClient` and so needs the same `onFocusReceived` override. The compiler stops at the first missing implementation, which is why the second class isn't visible until the first is patched.

Opened #1208 with the fix applied to both classes, scoped to just the Flutter 3.44 compat change (no QuillDeltaEncoder bundling), in case it's easier for maintainers to take that path. Happy to close it if you'd prefer to extend this PR.

rodolfogoulart:
Thank man.
So, for the other fix, I wait for the merge and create another pull request?

salihagic:
No need for a second PR — #1208 already includes the `NonDeltaTextInputService` override alongside the `DeltaTextInputService` one, so the whole Flutter 3.44 compat side is covered there.

Two ways the maintainers could handle it:

1. Merge #1208 for the Flutter 3.44 fix, and you keep this PR scoped to the QuillDeltaEncoder alignment work (which is unrelated and probably easier to review separately).
2. Or, if you'd rather keep everything in one PR, add the same `onFocusReceived` override to `NonDeltaTextInputService` here and I'll close #1208.

Either works for me — happy to follow whichever path the maintainers prefer.

PR #1230, opened 2026-08-06
fix: preserve mobile IME composing range and unstick hardware key handling

wldnjs990:
## Problem

Two related mobile IME bugs make CJK (Korean/Chinese/Japanese) typing unusable on Android.

**1. Composition is split at syllable boundaries.**
`_updateComposing` merges the previous composing range with the incoming delta's range. On mobile this corrupts the range at syllable boundaries — e.g. typing `ㄱ` after `반` yields `(0, 2)` instead of the keyboard-reported `(1, 2)`. `attach()` then sees a mismatched editing state, calls `setEditingState`, and the framework restarts the input method mid-composition, splitting the jamo apart.

**2. Backspace stops responding after a composition timeout.**
The IME signals the end of a composition (Android's composition timeout) with a `TextEditingDeltaNonTextUpdate`, which doesn't change the selection. `enableIMEShortcuts` is only refreshed on attach/selection changes, so it stays stale-`false` and `_onKeyEvent` swallows subsequent hardware key events. Backspace does nothing until the user taps the editor again.

### Steps to reproduce (Android)

1. Open the editor with a Korean keyboard.
2. Type `반`, then `ㄱ` — the composition splits instead of forming `방`.
3. Stop typing until the composition times out, then press backspace on a hardware keyboard — nothing happens until you tap the editor.

## Fix

- **`non_delta_input_service.dart`** — on mobile, trust the platform's composing range verbatim instead of merging it with the previous range. Mobile IMEs report the authoritative composing range on every delta. Desktop keeps the existing merge behaviour, so the macOS Chinese IME workaround is untouched.
- **`keyboard_service_widget.dart`** — re-sync `enableIMEShortcuts` with the actual composing state before gating key events, so a composition that ended via `TextEditingDeltaNonTextUpdate` no longer leaves the flag stale.

The second fix depends on the composing range being accurate on mobile, which the first one guarantees.

## Testing

Verified on a physical Android device: zero input-method restarts during composition, correct syllable composition, and backspace responsive after a composition timeout. Desktop behaviour is unchanged — the merge path is only skipped when `PlatformExtension.isMobile`.

## Scope

+19 lines across 2 files, no public API change.

### DIFF
--- a/lib/src/editor/editor_component/service/ime/non_delta_input_service.dart
+++ b/lib/src/editor/editor_component/service/ime/non_delta_input_service.dart
@@ -305,6 +305,14 @@ class NonDeltaTextInputService extends TextInputService with TextInputClient {
   void _updateComposing(TextEditingDelta delta) {
     if (delta is TextEditingDeltaNonTextUpdate) {
       composingTextRange = delta.composing;
+    } else if (PlatformExtension.isMobile) {
+      // Mobile IMEs (including composing languages such as Korean) report the
+      // authoritative composing range on every delta. Merging it with the
+      // previous range corrupts the range at syllable boundaries (e.g. 반+ㄱ
+      // becomes (0,2) instead of (1,2)), which makes attach() push a
+      // mismatched editing state and restart the IME mid-composition.
+      // Trust the platform value verbatim.
+      composingTextRange = delta.composing;
     } else {
       composingTextRange = composingTextRange != null &&
               composingTextRange!.start != -1 &&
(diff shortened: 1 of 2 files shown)


wldnjs990:
Hi maintainers, and thanks for all the work you put into this project.

I wanted to gently follow up on this PR. The CLA is signed and all
checks are green, with no conflicts against main.

For context on why it might be worth a look: it fixes two Android IME
issues that make Korean/Japanese/Chinese input unusable in practice.
The composition splits at syllable boundaries while typing, and
backspace stops responding after a composition timeout. Both are
verified on a physical Android device. The change is +19 lines across
2 files, with no public API change and no effect on desktop behavior.

I completely understand the review queue is long, so there's no rush
at all. Whenever someone has the bandwidth, I'd be glad to rebase,
split the commits, or add tests — whatever makes it easier to review.

Thanks again!

